PR #4, opened 2019-11-21
Contract invariant example

leonardoalt:
Consider the following contract written in Solidity:
```
contract C {
    uint x;
  
    function f() public {      
        if (x == 0)
            x = 1;             
    }

    function g() public {
        if (x == 1)
            x = 2;
    }

    function h() public {
        if (x == 2)
            x = 0;
    }
}
```

The added specification is an example of how to specify properties of this contract, with some things to discuss.

1. In this high level (no EVM) version of the specification language, I would suggest listing a contract's state variables in a `storage` section under a `contract`, where the types would match the ABI specification.
2. The state variables can then be used by the following function specifications without the need to specify a storage slot.
3. Contract invariants would be described in `invariants` also under `contract`. A property may hold over all functions, where the set of functions it holds for is not specified, or over a specified set of functions.
4. Each function can specify its post condition under `ensures`. Is this the same as the currently used `such that`?

### DIFF
--- a/examples/transitions/invariant.md
+++ b/examples/transitions/invariant.md
@@ -3,8 +3,6 @@
 ```act
 contract A
 
-storage
-
 	x : uint256
 
 invariants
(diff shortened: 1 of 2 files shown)


leonardoalt:
@MrChico can we start a discussion here and only merge when we agree on how it should look like ideally?

MrChico:
I mostly agree to all points, with some reservations:
1. The storage abi would have to be given somewhere, I agree, but they could possibly be extracted from the source as well with the new storage layout flag.
2. Both syntaxes are fine with me. I still think there's a place for the verbose syntax, but I can understand that the concise has it's qualities.
3. Agree.
4. Yes, `such that` and `ensures` are synonyms. I prefer `such that` to be in line with the other "english-like" declarators. 
That being said, what do you think of the distinction between post-state and postcondition? I would argue that the poststate is what is on the RHS of the storage header + return header, whereas the `such that` header would more be used to specify that a particular condition wrt these variables hold. In other words, I think your spec should look like:

```act
behaviour f of C
interface f()

storage

x |-> 0 => 1
```

Notice that using the verbose syntax, one also gets a distinction between pre-state and precondition. In the concise syntax, the corresponding would be:

```act
behaviour f of C
interface f()

storage 
x := 1

if 
x == 0
```

leonardoalt:
1. Right, it could be type checked against the ABI.
2. I think so too, but I would personally reserve that for EVM specific specs. When writing high level specs I wouldn't really care about the slot/offset.
4. Right, agree with the distinction.

PR #8, opened 2019-11-26
Add Solidity standard-json to Act skeleton translator

leonardoalt:
This simple script takes the output of a Solidity compilation with `storageLayout` and `abi` requested and builds an Act specification skeleton for the given contracts.

### DIFF
--- a/tools/act.py
+++ b/tools/act.py
@@ -1,0 +1,57 @@
+class Act:
+	"""This class is a text representation of an Act specification for a smart contract."""
+
+	def __init__(self, contract, contractName, path):
+		"""Builds the specification text from a JSON object.
+
+		Parameters
+		----------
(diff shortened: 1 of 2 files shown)


tools/act.py
```diff
@@ -0,0 +1,57 @@
...
+		"""Builds the specification of a function, including property and behavior."
+
+		Parameters
+		----------
+		function : JSON object
+			The function object from Solidity's ABI output.
+		"""
+		name = function["name"]
+		fProperty = "property " + name + ".post of " + name + "\n"
+		fBehavior = "behavior " + name + ".behavior of " + name + "\n"
```
MrChico: second name should be `contractName`

leonardoalt: @MrChico updated

Issue #5, opened 2019-11-25
If vs Case: globalizing pre- & postconditions


MrChico:
In the current version of `act`, the two headers `if` and `iff` have the following semantics:

The `if` header specifies preconditions that need to be met in order for the specification to be defined. Whatever happens in the case where the preconditions are not met is considered outside the scope of the current specification and must be defined elsewhere in order for the specification to be complete.

The `iff` header on the other hand, asserts that the function at hand will behave according to the specification when the conditions are met, but also asserts that it must REVERT otherwise. In other words, specifications with an `iff` header actually result in two proof obligations, a `pass` case where the assumptions are met, and a `fail` case where they are not. 

In images:
<img width="737" alt="Screen Shot 2019-11-25 at 17 21 36" src="https://user-images.githubusercontent.com/8930559/69558208-1fc2bd80-0fa8-11ea-875b-3398c920d576.png">

Here is an alternative approach which roughly changes the `if` header to a `case` pattern, allowing the remaining headers to still apply across all cases. Here's a conceptual example:
```act
behaviour transfer of Token
interface transfer(uint256 value, address to)

iff in range uint256

     balanceOf[CALLER] - value

iff

     CALLVALUE == 0
     to =/= CALLER => balanceOf[to] + value < 2^256

case to =/= CALLER:
storage
     balanceOf[CALLER] := balanceOf[CALLER] - value
     balanceOf[to]     := balanceOf[to] + value

case _:

     noop

end
```

MrChico:
Structuring specifications in this way has a couple of implications:
- The `iff` header is now global across all cases. If one wants to specify that a particular case actually reverts, we need a new keyword `revert`, as in:
```
function f(uint x) public {
if (x == 42) {
    revert();
   }
}
```

```act
behaviour f of A
interface f(uint x)

case x == 42:
    revert

case _:
    noop
```

– The postconditions (`ensures` header) would also be global. This suggests a particular way of thinking about smart contract methods: if one of your contract functions can leave the contract state in ways that would need to be expressed as different postconditions, then they should have really just been two different functions in the first place!

While this might seem like it would be restricting the features of the specification language, this is only true if we assert that there must be only one `behaviour` per `interface`. Without this assumption, this new suggestion is really just a new syntactic convention. 

I do suggest that we continue along with the current behaviour and do not impose this assumption. Function selectors are really just a source-level concept anyway, the contract really only has one method.

MrChico:
[...]

PR #11, opened 2020-01-22
Syntax updates

MrChico:
Contains three extensions to the basic syntax:
- Cases
- Multiple storage headers, for specifying changes to more than one contract
- Creation syntax, which includes an initial specification of the storage layout of the contract. This essentially creates the base case for the inductive system that defines the contract.

This answers the question of "where to declare the storage layout" of a contract by requiring them to be defined in the specification of the constructor.


MrChico:
Related to https://github.com/ethereum/act/issues/5

tests/token/transfer.act
```diff
@@ -7,18 +7,18 @@ iff
    value <= balanceOf[CALLER]
    CALLER =/= to => balanceOf[to] + value < 2^256
 
-storage
-
 case CALLER =/= to:
 
-   balanceOf[CALLER] => balanceOf[CALLER] - value
-   balanceOf[to]     => balanceOf[to] + value
+   storage
```
leonardoalt: Wouldn't this be confused with the storage layout of the contract?

examples/contractCreation/modest.act
```diff
@@ -0,0 +1,18 @@
+// a modest example specifying the creation of a new contract,
+// in this case the contract subject at question.
+// Speciying the creation of a contract serves multiple purposes:
```
leonardoalt: `Speciying` typo

examples/contractCreation/modest.act
```diff
@@ -0,0 +1,18 @@
+// a modest example specifying the creation of a new contract,
```
leonardoalt: `A modest...`

examples/contractCreation/modest.act
```diff
@@ -0,0 +1,18 @@
...
+// perspective, it defines the storage layout of the contract, by explicit
+// typing of storage variables.
+
+behaviour creation of Modest
+interface constructor()
+
+creates storage
+    // storage variables must be given an explicit initial value
+    // if they are not assigned a value in the constructor,
+    // this is 0. TODO (is this annoying?)
```
leonardoalt: Maybe we should be explicit and require one and only one assignment?

examples/multipleContracts/multi.act
```diff
@@ -0,0 +1,29 @@
+behaviour remote of B
+interface set_remote(uint _x)
+
+iff
+   CALLVALUE == 0
+
+storage
```
leonardoalt: Also, maybe we should call this `state` instead of `storage`? Storage sounds more low level to me. What do you think?

examples/multipleContracts/multi.act
```diff
@@ -0,0 +1,29 @@
...
+   // newAddr should be a builtin function
+   a => newAddr(ADDRESS, NONCE)
+
+// storage labels can also introduce new contracts
+creates storage newAddr(ADDRESS, NONCE)
```
leonardoalt: I don't really understand this section.

examples/multipleContracts/multi.act
```diff
@@ -0,0 +1,29 @@
+behaviour remote of B
+interface set_remote(uint _x)
+
+iff
+   CALLVALUE == 0
+
+storage
+
+  y => 1
```
leonardoalt: Where's `y` coming from?

tests/multi/multi.act
```diff
@@ -0,0 +1,24 @@
+behaviour remote of B
+interface set_remote(uint y)
+
+iff
+   CALLVALUE == 0
+
+// we can specify how a call to a function of B
+// updates the storage of anothere by labelling the storage header
```
leonardoalt: Typo `another`

leonardoalt: Wouldn't this idea of specifying how `a`'s storage in `B`'s spec get too convoluted at some point? IMO it would be modular so that `B`'s spec only knows that a certain function from `a` is called. By analyzing that function in `A` we could infer what happens to `B.a`.

tests/creation/create.act
```diff
@@ -0,0 +1,18 @@
+// a modest example specifying the creation of a new contract,
```
leonardoalt: `A modest...`

tests/creation/create.act
```diff
@@ -0,0 +1,18 @@
...
+// in this case the contract subject at question.
+// Speciying the creation of a contract serves multiple purposes:
```
leonardoalt: Typo `specifying`

tests/token/transfer.act
MrChico: it seems different in form enough to me to not cause any confusion. But calling it `state` would also alleviate this concern

examples/multipleContracts/multi.act
MrChico: the syntax `storage A` allows us to specify the state changes in the contract A in a spec of B. The syntax `creates storage` allows us to specify the initial state of a contract. This is the combination of the two

tests/multi/multi.act
MrChico: Hmm, I don't think I like that sort of indirection in specs. It would make it harder to see what's actually going on. And it would only work for contracts which has specs already. Yes, contracts that touch a lot of other stuff would have large specs, but that does tell you something about their complexity

examples/multipleContracts/multi.act
MrChico: `state` to me would also contain the contract `nonce` and `balance`

tests/multi/multi.act
MrChico: Thinking about this more, I think I can express this "indirection" concern in a more precise way. Since specs specify state transitions with pre and post state, composing several state transitions to make a larger one would require a careful ordering of them. 

It is not sufficient to say that `B` calls `a`, we need to say at what point this happens.

If I have a state transition `f` that calls `g` (for which we have a spec) somewhere in the middle, I would need to break `f` down into `f0`, which specifies how state changes prior to `g` being called and `f1` which specifies how the state changes afterwards. Then I could express `f` as `f0 ~> g ~> f1`. 

This composition is definitely useful, but I don't think that `behaviours` would be the right place for it. Behaviours, in my view, should specify the state transition as a pre-post pair. Combining state transitions can happen in a property section, or in other backends

examples/contractCreation/modest.act
MrChico: yup. One assignment only!

MrChico:
Thanks for the review. This should now be addressed

examples/multipleContracts/multi.act
leonardoalt: Is `A` missing there? Or do I still not get it?
`creates storage A newAddr(ADDRESS, NONCE)`

tests/multi/multi.act
leonardoalt: I guess "inlining" the state updates would remove the need for this split into `f0` and `f1`.

examples/multipleContracts/multi.act
MrChico: Yeah `creates storage A at newAddr(ADDRESS, NONCE)` might be necessary since we need to know the code for `A`.

tests/multi/multi.act
MrChico: How would that look? Currently state updates aren't specified sequentially so I don't see where it could be squeezed in

PR #17, opened 2020-04-24
Happy + Lex for parsing

MrChico:
This replaces the current BNFC set up with a Lex + Happy setup. 
All the examples now parse, and all but two typecheck and can be output in the IR format. Obviously more to do but I'd like to get this merged before it gets too large to review.


default.nix
```diff
@@ -1,2 +1,20 @@
 {nixpkgs ? import <nixpkgs> {}, compiler ? "ghc865"}:
-nixpkgs.pkgs.haskell.packages.${compiler}.callPackage (import ./src/default.nix) {}
+let
+  dapptools = builtins.fetchGit {
```
leonardoalt, line 3: Why is this needed?

src/Lex.x
```diff
@@ -0,0 +1,201 @@
...
+$space = [\ \t\f\v\r]
+
+tokens :-
+
+  -- ($space* \n)+       { mk BREAK }
+  -- $space+             ;
+  $white+                               ;
+
+  -- reserved words
+  behaviour                             { mk BEHAVIOUR }
```
leonardoalt, line 20: Can we also have a
`behavior { mk BEHAVIOUR }`
? ;)

src/Lex.x
```diff
@@ -0,0 +1,201 @@
...
+
+  -- builtin functions
+  newAddr                        { mk NEWADDR }
+
+  -- environment vars
+  CALLER                                { mk CALLER }
+  CALLVALUE                             { mk CALLVALUE }
+  ORIGIN                                { mk ORIGIN }
+  THIS                                  { mk THIS }  -- normally called address, but that's taken
+  NONCE                                 { mk NONCE } -- technically not an opcode
```
leonardoalt, line 63: Maybe also add block info?

src/Lex.x
```diff
@@ -0,0 +1,201 @@
...
+  "="                                   { mk EQ }
+  ">"                                   { mk GT }
+  "<"                                   { mk LT }
+  ":"                                   { mk COLON }
+  "+"                                   { mk PLUS }
+  "-"                                   { mk MINUS }
+  "*"                                   { mk STAR }
+  "/"                                   { mk SLASH }
+  "%"                                   { mk MOD }
+  "^"                                   { mk CARROT }
```
leonardoalt, line 88:
```suggestion
  "^"                                   { mk CARET }
```
Unsure if meme

src/Parse.y
```diff
@@ -0,0 +1,285 @@
...
+%monad { Err } { (>>=) } { return }
+%tokentype { Lexeme }
+%error { parseError }
+
+%token
+
+  eof                         { L EOF _ }
+
+  -- reserved words
+  'behaviour'                 { L BEHAVIOUR _ }
```
leonardoalt, line 20: Same as above, can we add `behavior`?

src/test/a.act
```diff
@@ -0,0 +1,12 @@
+behaviour add of SafeAdd
```
leonardoalt, line 1: I think it's better if we make the filenames more clear. Maybe `add.act`?

default.nix
MrChico: The nix packages for `hevm` and `sbv` are broken in `nixpkgs` so we need to manually override them right now

src/test/a.act
MrChico: sorry, this file shouldn't be here. SafeAdd is already in `tests/safemath`. Will remove this

MrChico:
Thanks @leonardoalt , your comments should now be addressed

src/Lex.x
```diff
@@ -59,6 +60,13 @@ tokens :-
   CALLER                                { mk CALLER }
   CALLVALUE                             { mk CALLVALUE }
   ORIGIN                                { mk ORIGIN }
+  BLOCKHASH                             { mk BLOCKHASH }
+  NUMBER                                { mk NUMBER }
```
leonardoalt, line 64: `BLOCKNUMBER` might be less confusing

default.nix
leonardoalt: Ok!

src/Lex.x
MrChico: yeah, I suppose we can be a bit foreboding of that eip..

MrChico:
@leonardoalt good to merge?

PR #19, opened 2020-05-11
Kbackend

MrChico:
Time to merge some work again!

This PR mainly deals with exporting act specs to the k specs, done in the Splitter.hs module, but also includes some internal simplifications and refactorings in typechecking and parsing.
Besides translating expressions into a form that k can parse, this also involves fetching storage locations from the `storage-layout` of a soljson file (in the future, this should be made optional so that manual storage locations can be given). 

There's still much to do, but I have now achieved the milestone of successfully proving `Token_transfer` using [this branch](https://github.com/dapphub/klab/tree/act2) of klab!


README.md
```diff
@@ -130,6 +130,19 @@ We are currently working on three different proof backends:
 	- SMT theorems
 
 
+The types of proofs that can be performed by the various backend can be understood in the following table:
```
leonardoalt, line 133:
```suggestion
The types of proofs that can be performed by the various backends can be understood in the following table:
```

README.md
```diff
@@ -130,6 +130,19 @@ We are currently working on three different proof backends:
...
+
+| Proof type \ Backend          | HEVM    | KEVM | SMT     | Coq     |     |
+| ----------------------------- | ------- | ---- | ------- | ------- | --- |
+| Bytecode (with unknown calls) | Planned | ?    |         |         |     |
+| Bytecode                      | Planned | WIP  |         |         |     |
+| Poststate => postcondition    |         |      | Planned |         |     |
+| Source level                  |         |      | Planned |         |     |
+| Contract level invariant      |         |      |         | Planned |     |
```
leonardoalt, line 141:
```suggestion
| Contract level invariant      |         |      | Planned | Planned |     |
```
?

leonardoalt: What's the rightmost column?

src/Lex.x
```diff
@@ -21,7 +21,6 @@ tokens :-
   behavior                              { mk BEHAVIOUR }
   of                                    { mk OF }
   interface                             { mk INTERFACE }
-  constructor                           { mk CONSTRUCTOR }
```
leonardoalt, line 24: Why was constructor removed?

src/Main.hs
```diff
@@ -334,21 +340,49 @@ inferExpr env exp = let intintint op v1 v2 = do w1 <- checkInt env v1
...
     EGT _   v1 v2 -> boolintint  GE   v1 v2
-    ETrue _ ->  Ok  $ ExpBool $ LitBool True
-    EFalse _ -> Ok $ ExpBool $ LitBool False
-    -- EITE _ v1 v2 v3 -> do w1 <- checkBool env v1
-    --                       w2 <- checkInt env v2
-    --                       w3 <- checkInt env v3
-    --                       Ok $ ITE w1 w2 w3
+    ETrue _ ->  Ok . ExpBool $ LitBool True
+    EFalse _ -> Ok . ExpBool $ LitBool False
+    -- TODO: make ITE polymorphic
```
leonardoalt, line 345: In this PR or later?

src/Splitter.hs
```diff
@@ -1,56 +1,310 @@
...
--- kaseSplit :: Behaviour -> SolcContract -> (VM, Signature, Precondition, Postcondition)
--- kaseSplit = error "TODO"
+
+import EVM.Solidity (SolcContract(..), StorageItem(..), SlotType(..))
+import Control.Monad
+import Data.Map.Strict (Map) -- abandon in favor of [(a,b)]?
+import qualified Data.Map.Strict as Map -- abandon in favor of [(a,b)]?
+
+-- Transforms a RefinedSyntax.Behaviour
+-- to a k spec.
```
leonardoalt, line 29: Maybe the file name could represent that better?

tests/smoke/smoke.act.parsed.hs
```diff
@@ -1 +1 @@
-[Transition "f" "A" f(AbiUIntType 256 x) [] (Branch (AlexPn 0 0 0) (BoolLit True) [Leaf (AlexPn 0 0 0) (BoolLit True) (Post Nothing [] (Just (IntLit 1)))]) Nothing]
+[Transition "f" "A" f(uint256 x) [] (Branch (AlexPn 0 0 0) (BoolLit True) [Leaf (AlexPn 0 0 0) (BoolLit True) (Post Nothing [] (Just (IntLit 1)))]) Nothing]
```
leonardoalt, line 1: Nice!

src/Lex.x
MrChico: TL;DR: to unify identifiers and the `constructor` keyword

I didn't want to make constructor a reserved word because I wanted to be able to name constructor specs as for example
```
behaviour constructor of A
interface constructor()

...
```

but also as:
```
behaviour init of A
interface constructor()

...

```

which meant that I would need to have two different parsing rules, (either taking a `constructor` or an id), or just forget about `constructor` having a special meaning in the first place. It doesn't matter what you end up calling the interface that defines the constructor, what matters is only the arguments.

I've also been toying with the idea of calling initialization specs something different than `behaviour` altogether, but haven't found a good alternative yet.

README.md
MrChico: hmmm, a mistake I think

MrChico: I think the right most column is supposed to be coq, and the penultimate column from the right is not supposed to exist

src/Splitter.hs
MrChico: Yeah, the file name should change. I was afraid git would screw up the diff if I did this in the same PR. Will go in the next one!

src/Main.hs
MrChico: Later – there are some problems with it I haven't quite figured out yet

src/Splitter.hs
leonardoalt: Ok!

src/Lex.x
leonardoalt: Ok!

README.md
leonardoalt: Wait, it still has this rightmost column with empty header

MrChico: yes, sorry, this shouldn't be marked as resolved

PR #20, opened 2020-06-08
Add binary search example

leonardoalt:


examples/binarySearch/binarySearch.act
```diff
@@ -0,0 +1,36 @@
+behaviour init of BinarySearch
+interface constructor()
+
+creates
+	uint256[] array
```
MrChico, line 5:
```suggestion
	uint256[] array := []
```

examples/binarySearch/binarySearch.act
```diff
@@ -0,0 +1,36 @@
...
+case _:
+	returns binarySearch(array, elem, 0, array.length)
+
+// postcondition
+// if elem in array, i | array[i] == elem
+// if elem not in array, -1
+
+where
+
+sorted(uint256[] a) =
```
MrChico, line 24: I'd place this definition under the `init` behaviour

examples/binarySearch/binarySearch.act
```diff
@@ -0,0 +1,36 @@
...
+invariants
+	sorted(array)
+
+// Returns the index of elem if elem in array,
+// otherwise returns -1.
+behaviour search of BinarySearch
+interface search(uint256 elem)
+
+case _:
+	returns binarySearch(array, elem, 0, array.length)
```
MrChico, line 16:
```suggestion
returns binarySearch(array, elem, 0, array.length)
```

leonardoalt:
Fixed the comments, but I didn't change the `binarySearch` function to do sequential pattern matching because I'm not sure it would make sense here.

leonardoalt:
Taking out of draft mode for review.
@MrChico I rewrote locally the `sorted` property in the functional style, but tbh I kinda prefer the quantified formula.
What do you think?

MrChico:
I'm still a little on the fence about having that quantifier in the definition. While it has a direct semantics in an smt backend, it would not be as easily translated into other settings.

In a type theoretic backend, in order to interpret this, the type of `a` would not be a list, but rather the pair `(Nat, Nat -> (Nat + ())`, for length and a "lookup" function. I'm not sure this the way we want to represent lists in that setting

leonardoalt:
Ok, fair enough.
The function does seem to be more general, will go with that.

examples/binarySearch/binarySearch.act
```diff
@@ -0,0 +1,38 @@
...
+// Returns the index of elem if elem in array,
+// otherwise returns -1.
+behaviour search of BinarySearch
+interface search(uint256 elem)
+
+returns binarySearch(array, elem, 0, array.length)
+
+// postcondition
+// if elem in array, i | array[i] == elem
+// if elem not in array, -1
```
leonardoalt, line 27: @MrChico what do you think about this /\?

examples/binarySearch/binarySearch.act
```diff
@@ -0,0 +1,40 @@
...
+
+invariants
+	sorted(array)
+
+where
+
+sorted (uint256[] array)
+case [] = true
+case [x] = true
+case [x,y,xs] = x <= y and sorted([y,xs])
```
MrChico, line 15: how about:
```
sorted : uint256[] -> bool
sorted ([]) = true
sorted ([x]) = true
sorted ([x;y;xs]) = x <= y and sorted([y;xs])
```

MrChico: possibly without the parens

examples/binarySearch/binarySearch.act
MrChico: I would write it like this, I think:
```
ensures

if elem in array then binarySearch(array, elem, 0, array.length) == elem else binarySearch(array, elem, 0, array.length) == -1
```
```

Comment (MrChico) [reply examples/binarySearch/binarySearch.act:27]: or, if you want to refer to `i` here, add `i := binarySeach(array, elem, 0, array.length)` to the where section

File: examples/binarySearch/binarySearch.act (hunk 5 above)
Comment (leonardoalt) [reply examples/binarySearch/binarySearch.act:15]: sounds good too

Comment (leonardoalt) [reply examples/binarySearch/binarySearch.act:15]: why `;` instead of `,`?

File: examples/binarySearch/binarySearch.act (hunk 4 above)
Comment (leonardoalt) [reply examples/binarySearch/binarySearch.act:27]: ok, so we're going with `e in c` for `e is a member of container c`? Did we have that syntax already?

File: examples/binarySearch/binarySearch.act (hunk 5 above)
Comment (MrChico) [reply examples/binarySearch/binarySearch.act:15]: I wanted to distinguish between `(,) : a -> a -> [a]` and `(;) -> a -> [a] -> [a]`.
Another choice is `::`, which idris uses for example

Comment (MrChico) [reply examples/binarySearch/binarySearch.act:15]: coq uses `;`. I think I prefer `::`

File: examples/binarySearch/binarySearch.act (hunk 4 above)
Comment (MrChico) [reply examples/binarySearch/binarySearch.act:27]: I don't think we introduce it yet but I like it!

Comment (leonardoalt) [reply examples/binarySearch/binarySearch.act:27]: updated

File: examples/binarySearch/binarySearch.act (hunk 5 above)
Comment (leonardoalt) [reply examples/binarySearch/binarySearch.act]: updated

File: examples/binarySearch/binarySearch.act [hunk 6]
```diff
@@ -0,0 +1,44 @@
...
+
+ensures
+
+if elem in array
+then array[index] = elem
+else index = -1
+
+where
+
+index = binarySearch(array, elem, 0, array.length)
```
MrChico, line 33:
```suggestion
index := binarySearch(array, elem, 0, array.length)
```

MrChico: at least that's what we've been doing so far. Maybe it's tedious...

MrChico: but it would leave `=` as comparison, the way you are using it above

MrChico: although we've been using `==` in other examples so far

examples/binarySearch/binarySearch.act
```diff
@@ -0,0 +1,44 @@
...
+behaviour search of BinarySearch
+interface search(uint256 elem)
+
+returns binarySearch(array, elem, 0, array.length)
+
+ensures
+
+if elem in array
+then array[index] = elem
+else index = -1
```
MrChico, line 29:
```suggestion
then array[index] == elem
else index == -1
```

examples/binarySearch/binarySearch.act
```diff
@@ -0,0 +1,44 @@
...
+
+where
+
+index = binarySearch(array, elem, 0, array.length)
+
+binarySearch(uint256[] a, uint256 elem, uint256 left, uint256 right) =
```
MrChico, line 35:
```suggestion
binarySearch(uint256[] a, uint256 elem, uint256 left, uint256 right) :=
```

examples/binarySearch/binarySearch.act
leonardoalt: updated `index` and `binarySearch`

examples/binarySearch/binarySearch.act
leonardoalt: `sorted` uses `=`, so `=` or `:=`?

MrChico: let's go with `:=` for now. We are using `:=` for assignment in storage declarations right now so this would be in line with that

leonardoalt: updated

examples/binarySearch/binarySearch.act
```diff
@@ -0,0 +1,45 @@
...
+if elem in array
+then array[index] == elem
+else index == -1
+
+where
+
+index := binarySearch(array, elem, 0, array.length)
+
+binarySearch : uint256[] -> uint256 -> uint256 -> uint256 -> uint256
+binarySearch(uint256[] a, uint256 elem, uint256 left, uint256 right) :=
```
dddejan, line 36: What guarantees termination of this function?

leonardoalt: That would need to be proven by `something`

dddejan: I thought this is the spec. Spec should somehow be self sufficient. If you start putting code into spec things will get complicated.

leonardoalt: This example is just something you _could_ do if you want the reasoning to be done also inside the act, but don't necessarily have to. The motivation was the deposit contract, where we wanted the proof backends to try to prove its properties.

PR #23, opened 2020-06-14
parser rework

endorphin:
fixing shift / reduce conflicts and trying to make things readable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+src/Parse.hs
+src/Lex.hs
 bin
 *.hi
 *.o
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,8 @@
 bin
+*.hi
+*.o
+*.dyn_hi
+*.dyn_o
+*.swp
+*~
+\#*\#
--- a/src/Lex.x
+++ b/src/Lex.x
@@ -1,6 +1,12 @@
 {
 
-module Lex (LEX (..), Lexeme (..), AlexPosn (..), lexer) where
+module Lex
+  ( Lex (..)
+  , Lexeme (..)
+  , AlexPosn (..)
+  , lexer
+  ) where
+
 import Prelude hiding (EQ, GT, LT)
 
 }
@@ -19,7 +25,7 @@ tokens :-
   -- reserved words
   behaviour                             { mk BEHAVIOUR }
   behavior                              { mk BEHAVIOUR }
-  constructor                           { mk CONSTRUCTOR }
+  creator                               { mk CREATOR }
   of                                    { mk OF }
   interface                             { mk INTERFACE }
   creates                               { mk CREATES }
@@ -30,7 +36,6 @@ tokens :-
   
   iff $white+ in $white+ range          { mk IFFINRANGE }
   iff                                   { mk IFF }
-  fi                                    { mk FI }
   and                                   { mk AND }
   or                                    { mk OR }
   not                                   { mk NOT }
@@ -109,14 +114,14 @@ tokens :-
 
 {
 
-data LEX =
+data Lex =
 
     BREAK
   | EOF
 
   -- reserved words
   | BEHAVIOUR
-  | CONSTRUCTOR
+  | CREATOR
   | OF       
   | INTERFACE
   | CREATES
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -1 +1,8 @@
-bin
\ No newline at end of file
+bin
+*.hi
+*.o
+*.dyn_hi
+*.dyn_o
+*.swp
+*~
+\#*\#
```
MrChico, line 8: you might be interested in this https://stackoverflow.com/questions/151945/how-do-i-control-how-emacs-makes-backup-files

src/Lex.x
```diff
@@ -12,13 +14,12 @@ $space = [\ \t\f\v\r]
...
 tokens :-
 
-  -- ($space* \n)+       { mk BREAK }
-  -- $space+             ;
   $white+                               ;
 
   -- reserved words
   behaviour                             { mk BEHAVIOUR }
   behavior                              { mk BEHAVIOUR }
+  constructor                           { mk CONSTRUCTOR }
```
MrChico, line 22: I had this lexeme before, but I removed it because I didn't like how it made me unable to write specs:
```
behaviour constructor of A
interface contructor()
```
and I would have to make up some other name for this behaviour. One could also skip its name and just call it a constructor of course, but that's why I removed it before

src/Parse.y
```diff
@@ -1,16 +1,23 @@
...
 import EVM.Solidity (SlotType(..))
 import qualified Data.List.NonEmpty as NonEmpty
 import Syntax
-import ErrM
+
 }
 
 %name parse
-%monad { Err } { (>>=) } { return }
+%monad { Either String } { (>>=) } { return }
```
MrChico, line 20: the `Err` monad carries position information and not just a string, which gives us some more options for more comprehensive error messages

src/RefinedAst.hs
```diff
@@ -1,54 +1,53 @@
...
   | Boolean
   | ByteStr
---  | Mapping (Map MType MType)
   deriving (Eq, Ord, Show, Read)
 
---data T_List t
---data T_Tuple
-
+data StorageUpdate t where
+
```
MrChico, line 50: what a cliffhanger!

src/Syntax.hs
```diff
@@ -1,145 +1,135 @@
...
+  | ESub Expr Expr      -- `a - b`
+  | EMul Expr Expr      -- `a * b`
+  | EDiv Expr Expr      -- `a / b`
+  | EMod Expr Expr      -- `a % b`
+  | EExp Expr Expr      -- `a ^ b`
+
+  -- other
+  | ERead Ref           -- `a`
+  | EEnv EthEnv         -- `CALLVALUE`
+  | EITE Expr Expr Expr -- `if a then b else c`
```
MrChico, line 109: with `fi` or no? You added its lexeme. If `fi` can be avoided I would like to avoid it

src/Lex.x
endorphin: Hm, you're right we should find a different name because "constructor" is a common name. I do think behaviours should be syntactically distinguished from constructors though, since they have different semantics. Should be something like
```
constructor_keyword constructor of Contract
interface ...
```

src/Parse.y
endorphin: OK, will revert. I had removed position info temporarily to make it a bit easier to hack the grammar.

src/Syntax.hs
endorphin: Yeah, I wanted to ask you about this. It can be avoided. Without the `fi`, the ambiguous `if a then b else c + d` will always parse as `if a then b else (c + d)` rather than `(if a then b else c) + d`, which seems like a reasonable norm.

src/Parse.y
endorphin: Actually, why do we need to track position info in the error monad? `parseError` has access to the current position, regardless, by just looking at the remaining list of lexemes.

src/Lex.x
MrChico: we could also disallow custom naming of the `constructor` section at all and just do:
```
constructor of Contract
interface ...
```

src/Parse.y
MrChico: ok, so we could have a function `parseError -> AlexPosn` ?

src/Syntax.hs
MrChico: I agree

src/Lex.x
endorphin: i chose `creator` for now, and removed their name field

src/Parse.y
```diff
@@ -2,16 +2,20 @@
 
   this is a happy grammar :)
 
+  14 shift/reduce conflicts in a single state are expected, due
+  to the ambiguity of if/then/else expressions.
```
MrChico, line 6: can we get rid of them by saying they are left associative?

endorphin: no, since if/then/else isn't a binary operation. we could probably rewrite the grammar to be unambiguous but that would needlessly complicate things.

MrChico:
Many changes from here were incorporated into #33

Issue #24, opened 2020-06-20
hevm build errors in nix-shell


xwvvvvwx:
I'm getting some build errors in `hevm` when I run `nix-shell` from the project root: 

```
Setup: Missing dependency on a foreign library:
* Missing (or bad) C library: ff
```

full log: https://ipfs.io/ipfs/QmQ7dYfuhzQtJSNAssx2EtszXxyMmWaeqLPHTVfFKHnYaA/out.log

Any idea whats going on?

endorphin:
I'm not sure, although it looks like this has come up before:

https://github.com/dapphub/dapptools/issues/223

If you haven't already, check that you're on a relatively recent nixpkgs (I just checked that it builds for me on the current 19.09 stable channel).

Otherwise, I have occaisionally had permissions issues with C library files in `/nix/store` needed by ocaml libs, so it could be something like that.

xwvvvvwx:
@endorphin I'm building with `nixpkgs` pinned in the `shell.nix` to `19.09@8260cd5bc65fad88ad8881b32c990e5659cc417a`, so I don't think that's the problem. :thinking:

MrChico:
Hmmm, it _should_ be configured here https://github.com/dapphub/dapptools/blob/91d7226f2f12453dbc867faa58db2a4342bea479/haskell.nix#L52

PR #25, opened 2020-07-09
Rework nix infra

xwvvvvwx:
This reworks the nix infra based on https://github.com/utdemir/hs-nix-template.

The rework has the following benefits:

- more reproducible (pinned `nixpkgs`)
- code sharing between `default.nix` and `shell.nix`
- manual calls to `cabal2nix` are no longer requried
- fixes https://github.com/ethereum/act/issues/24 on my machine
- support for `direnv` and `lorri`

A small fix to the broken import from `EVM.Symbolic` in `Splitter.hs` is also bundled with this PR.


MrChico:
Nice! Can `src/default.nix` also be removed by this PR?

xwvvvvwx:
yes! I didn't even notice that file :grin: :pray:

default.nix
```diff
@@ -1,20 +1,47 @@
-{nixpkgs ? import <nixpkgs> {}, compiler ? "ghc865"}:
+{ compiler ? "ghc865" }:
+
 let
-  dapptools = builtins.fetchGit {
-    url = "https://github.com/dapphub/dapptools.git";
-    rev = "c44db1f252b870c0913d0f83093347e9c05469c9";
+  # git ls-remote https://github.com/NixOS/nixpkgs-channels.git nixos-19.09
```
MrChico, line 4: any reason to not go latest here? I know dapptools is on 19.09 but if 20.03 works with the overrrides it might be preferable

xwvvvvwx: version conflicts: https://ipfs.io/ipfs/QmcrDwY3Y3DPT1t58qzhfG6XWUuKSw5Ajtq6RQ8Qf3qicG/log.txt

```
 This package indirectly depends on multiple versions of the same package. This is very likely to cause a compile failure.
```

xwvvvvwx: I guess that it's getting confused because `hevm` still uses `19.09` even when we bump to `20.03` here, which is causing multiple versions of the same package to be used?

MrChico: Gotcha. Yeah, using 19.09 seems good then

PR #30, opened 2020-08-11
Fix parse tests

MrChico:
Fixed ci!


Makefile
```diff
@@ -12,7 +12,7 @@ src/Lex.hs: src/Lex.x
 
 # builds the rest of the haskell files (compiler)
 bin/act: src/Lex.hs src/Parse.hs src/Main.hs
-	cd src && cabal v2-install --installdir=../bin --overwrite-policy=always && cd ..
+	cd src && cabal new-update && cabal v2-install --installdir=../bin --overwrite-policy=always && cd ..
```
xwvvvvwx, line 15: I don't really know how cabal works. Does this break purity somehow?

MrChico: not sure if it breaks purity.... 
It isn't installing any dependencies, only updating its index, so hopefully it should be alright 🤞

PR #32, opened 2020-08-13
Constructors

xwvvvvwx:
Minimal support for construcors in the typed output. Draft for visibility (I'm assuming I made a bunch of n00b haskell errors :grin:)

## TODO

- [x] expand support for types
- [x] propogate conditions for bounds on storage variables as preconditions to `Transition` behaviours
- [x] add `Invariants` type to refined AST and use this to hold invariants
- [x] tests


src/Main.hs
```diff
@@ -188,38 +188,93 @@ andRaw [] = BoolLit True
...
+    -- flatten case tree
+    flatten :: [Exp Bool] -> [Exp Bool] -> [Exp Bool] -> Case Expr Post -> Err [Behaviour]
+    flatten iff postc pathcond (Leaf _ cond post) = do
+      c <- checkBool env cond
+      (p, maybeReturn, contracts) <- checkPost env contract post
+      return $ splitCase name False contract iface (c : pathcond) iff maybeReturn p postc contracts
+    flatten iff postc pathcond (Branch _ cond cs) = do
+      c <- checkBool env cond
+      leaves <- mapM (flatten iff postc (c : pathcond)) (normalize cs)
+      return $ join leaves
```
xwvvvvwx, line 221: Most of this was just adding type signatures and  reformatted it for line width.

The only real change was to pull `splitCase` out so it could be shared between the two pattern matches.

src/Main.hs
```diff
@@ -188,38 +188,93 @@ andRaw [] = BoolLit True
...
+
+  return $ splitCase name True contract iface [] iffs' Nothing stateUpdates postcs [contract]
+  where
+    env = (fromMaybe mempty (Map.lookup contract store), store, abiVars)
+    abiVars = Map.fromList $ map (\(Decl typ var) -> (var, metaType typ)) decls
+    bounds = (getStorageBounds <$> assigns)
+
+    -- computes the storage bounds from the types in an `Assign`
+    getStorageBounds :: Assign -> Exp Bool
+    getStorageBounds (AssignVal (StorageVar (StorageValue (AbiUIntType size)) id) _)
```
xwvvvvwx, line 241: probably this could be made nicer by using `metaType`

src/Main.hs
```diff
@@ -188,38 +188,93 @@ andRaw [] = BoolLit True
...
+    getStateUpdates (AssignVal (StorageVar (StorageValue (AbiUIntType size)) id) expr) = do
+      val <- checkInt env expr
+      return $ [(id, [Right (IntUpdate (DirectInt id) val)])]
+    getStateUpdates (AssignVal (StorageVar typ id) expr)
+      = error $ "todo: type" ++ show typ ++ "is unsupported in constructors"
+    getStateUpdates _ = error "todo: support multiple and struct assignment in constructors"
+
+    -- processes the expressions in the invariant block
+    processInvariants :: Ensures -> Err [Exp Bool]
+    processInvariants exprs = mapM (\e -> checkBool env e) exprs
```
MrChico, line 258:
```suggestion
    processInvariants exprs = mapM (checkBool env) exprs
```

src/Main.hs
```diff
@@ -188,38 +188,93 @@ andRaw [] = BoolLit True
...
+      = And (LE (LitInt 0) (IntVar id)) (LE (IntVar id) (Exp (LitInt 2) (LitInt $ toInteger size)))
+    getStorageBounds (AssignVal (StorageVar typ id) expr)
+      = error $ "todo: type" ++ show typ ++ "is unsupported in constructors"
+    getStorageBounds _ = error $ "todo: support multiple and struct assignment in constructors"
+
+    -- computes the state updates from an `Assign`
+    getStateUpdates :: Assign -> Err [(Id, [Either StorageLocation StorageUpdate])]
+    getStateUpdates (AssignVal (StorageVar (StorageValue (AbiUIntType size)) id) expr) = do
+      val <- checkInt env expr
+      return $ [(id, [Right (IntUpdate (DirectInt id) val)])]
```
MrChico, line 251:
```suggestion
      return [(id, [Right (IntUpdate (DirectInt id) val)])]
```

src/Main.hs
```diff
@@ -188,38 +188,93 @@ andRaw [] = BoolLit True
...
+  iffs' <- checkIffs env (iffs <> calldataBounds)
+
+  invariants <- processInvariants $ fromMaybe [] maybeInvariants
+  let postcs = bounds <> invariants
+
+  return $ splitCase name True contract iface [] iffs' Nothing stateUpdates postcs [contract]
+  where
+    env = (fromMaybe mempty (Map.lookup contract store), store, abiVars)
+    abiVars = Map.fromList $ map (\(Decl typ var) -> (var, metaType typ)) decls
+    bounds = (getStorageBounds <$> assigns)
```
MrChico, line 237:
```suggestion
    bounds = getStorageBounds <$> assigns
```

src/Main.hs
```diff
@@ -188,38 +188,93 @@ andRaw [] = BoolLit True
...
+      (p, maybeReturn, contracts) <- checkPost env contract post
+      return $ splitCase name False contract iface (c : pathcond) iff maybeReturn p postc contracts
+    flatten iff postc pathcond (Branch _ cond cs) = do
+      c <- checkBool env cond
+      leaves <- mapM (flatten iff postc (c : pathcond)) (normalize cs)
+      return $ join leaves
+
+splitBehaviour store (Constructor name contract iface@(Interface _ decls) iffs creates@(Creates assigns) extStorage maybeEnsures maybeInvariants) = do
+  rawUpdates <- mapM getStateUpdates assigns
+  let stateUpdates = Map.fromList $ concat $ rawUpdates
```
MrChico, line 225:
```suggestion
  let stateUpdates = Map.fromList $ concat rawUpdates
```

src/Main.hs
MrChico: Couöd be nice to lift out env and abivars similarly?

xwvvvvwx:
@MrChico I am wondering if it makes sense to store the invariants outside of the `Behanviour`?

Currently they are just added as post conditions in the constructor `Behaviour`, but this seems like it will make it annoying to figure out what the contract level invariants are supposed to be.

I thought about maybe a new type in the Refined AST (`Invariant` or similar). The various proof backends could handle this as they wish. Does this make sense?

I'm also wondering if the constructor declaration is the best place to declare an invariant syntactically. What do you think about adding a block `invariant of <contract>`? I saw you had a `property` block in the deposit contract example, would that make sense to use for this purpose or did you have something else in mind there?

MrChico:
They can definitely belong to a different type than `Behaviour` after parsing & typechecking. Making an `Invariant` type sounds very sensible. Whether they be distinguished on the front end side or not I don't really know what I prefer. 
One could imagine both variants being supported. If you prefer, we can put it in a separate section for now

xwvvvvwx:
What makes you think that invariants should be declared in the constructor?

MrChico:
I imagine that the first thing you would look at when learning about a contract would be its constructor to get information about what it keeps in storage. If this also maintains a set of invariants over those storage variables it could give you some insight into what sort of contract you are looking at.

src/RefinedAst.hs
```diff
@@ -13,24 +13,27 @@ import Data.Map.Strict    (Map)
...
-   _contract :: Id,
-   _interface :: Interface,
-   _preconditions :: Exp Bool,
-   _postconditions :: Exp Bool,
-   _contracts :: [Id], -- can maybe be removed; should be equivalent to Map.keys(_stateupdates)
-   _stateUpdates :: Map Id [Either StorageLocation StorageUpdate],
-   _returns :: Maybe ReturnExp
-  }
+data Behaviour =
+  Invariants [Exp Bool] [Id]
```
xwvvvvwx, line 23: @MrChico this is a bit messy because there is already an `Invariants` type in `Syntax.hs`. Do you have any suggestions for better naming / types here?

src/Main.hs
```diff
@@ -176,50 +176,152 @@ defaultStore =
...
+
+-- extracts bounds on Integer values in storage, returns Iff or Exp Bool
+-- representations for use in either pre or post conditions
+getStorageBounds :: Env -> ([Exp Bool], [IffH])
+getStorageBounds (ours, _, _) =
+  unzip $ catMaybes $ fmap getBound $ Map.toList ours
+  where
+    getBound :: (Id, SlotType) -> Maybe (Exp Bool, IffH)
+    getBound (id, (StorageValue typ)) = case metaType typ of
+      Integer -> Just $ (bound typ (IntVar id), IffIn nowhere typ [EntryExp nowhere id []])
```
xwvvvvwx, line 263: I'm not sure how to add bounds for non `Integer` types (e.g. `Bool`, `Bytestr`, mappings)? Would that even make sense?

src/Main.hs
```diff
@@ -176,50 +176,152 @@ defaultStore =
...
+    ByteStr -> do
+      val <- checkBytes env expr
+      return (id, [BytesUpdate (DirectBytes id) val])
+checkAssign env (AssignMany (StorageVar (StorageMapping (keyType :| _) valType) id) defns)
+  = do updates <- mapM (checkDefn env keyType valType id) defns
+       return $ (id, updates)
+checkAssign env (AssignVal (StorageVar (StorageMapping _ _) _) _)
+  = Bad (nowhere, "Cannot assign a single expression to a composite type")
+checkAssign env (AssignMany (StorageVar (StorageValue _) _) _)
+  = Bad (nowhere, "Cannot assign multiple values to an atomic type")
```
xwvvvvwx, line 297: Would it make sense to use a `Defn` with e.g. an array type?

src/RefinedAst.hs
```diff
@@ -126,30 +129,36 @@ data ReturnExp
 
 -- intermediate json output helpers ---
 instance ToJSON Behaviour where
-  toJSON (Behaviour {..}) = object  [ "name" .= _name
-                                    , "contract"  .= _contract
+  toJSON (Invariants exps contracts) = object [ "kind" .= (String "Invariants")
```
xwvvvvwx, line 132: I aded this "Kind" key to the JSON output to differentiate between `Behaviour` and `Invariants`, is this name OK for you @MrChico ?

xwvvvvwx:
@MrChico I'm pretty happy with this as it stands, and I think it should be enough to start on support for inductive invariants in the SMT backend.

xwvvvvwx:
> I imagine that the first thing you would look at when learning about a contract would be its constructor to get information about what it keeps in storage. If this also maintains a set of invariants over those storage variables it could give you some insight into what sort of contract you are looking at.

I guess that makes sense. I'm gonna leave it as it is for now in the interests of saving time to move forwards with the SMT backend.

xwvvvvwx:
@MrChico would it make sense to add the contents of the `Invariants` block as `ensures` in the generated K specs?

src/Main.hs
MrChico: I don't think such bounds are necessary. We will be modeling those using `Bool` and `ByteString` as "metatypes", i.e. in the proving backends, so they won't need to be further constrained.

src/RefinedAst.hs
MrChico: Hmm, I think we might wanna lift `Invariants` to their own type, and let `Behaviour` just consist of `Behaviour` here. They could both be contained in a larger type, like `data Claim = B Behaviour | I Invariant` or `data Spec = Function Behaviour | Contract Invariant` or something like this. I'd also probably do `data Invariant = Invariant Id Exp Bool`, so we can tackle different claims individually.

src/RefinedAst.hs
MrChico: kind is good! 👍

tests/creation/create.act.typed.json
```diff
@@ -1 +1,97 @@
...
+                0,
+                "x"
+              ],
+              "symbol": "<="
+            },
+            {
+              "arity": 2,
+              "args": [
+                "x",
+                1.157920892373162e+77
```
MrChico, line 56: hmm, we should probably print this as a string to avoid javascript doing nasty things to our numbers

src/Main.hs
MrChico: Hmm, I don't think we want to give arrays an index like that. In general, there's a few things missing to deal with arrays right now, I think they are not really a priority atm.

I'm thinking that the way to deal with them would be to add some array flavored expressions to `Expr`, such as `List [Expr]`, and corresponding zooming / lookup operations

MrChico:
> @MrChico would it make sense to add the contents of the `Invariants` block as `ensures` in the generated K specs?

Yeah, that would definitely make sense

src/RefinedAst.hs
```diff
@@ -8,32 +8,51 @@
...
+   _interface :: Interface,
+   _preconditions :: Exp Bool,
+   _postconditions :: Exp Bool,
+   _contracts :: [Id], -- can maybe be removed; should be equivalent to Map.keys(_stateupdates)
+   _stateUpdates :: Map Id [Either StorageLocation StorageUpdate],
+   _returns :: Maybe ReturnExp
+  }
+
+catInvs :: [Claim] -> [Invariant]
+catInvs = fmap unwrapI . Data.List.filter isI
```
MrChico, line 40:
```suggestion
catInvs = fmap unwrapI . filter isI
```

MrChico: another way.
```
 catInvs :: [Claim] -> [Invariant] 
catInvs [] = []
catInvs ((I i):claims) = i:(catInvs claims)
catInvs ((B _):claims) = catInvs claims
```
feel free to use whichever you prefer

xwvvvvwx: ah nice, total functions!!! thx :pray:

MrChico:
This looks good to me, but it looks like test vectors need updating again (which confuses me a little)

xwvvvvwx:
Yeah not sure what's going on with the tests, need to investigate. Also the generated K specs for the constructors seem to be missing some post conditions, so something seems off there as well.

xwvvvvwx:
K specs should be fixed with https://github.com/ethereum/act/pull/32/commits/cbe2778d0b61edeb6c9cdd367b3d7476feebd22d and the test vectors needed updating after the fix in https://github.com/ethereum/act/pull/32/commits/553f673a9a590a66579b4c679b32eb0d8472b635.

xwvvvvwx:
Not really related to this PR, but they were small fixes so I removed postconditions and state updates from the fail specs.

I'm ready to merge this if you are @MrChico...

MrChico:
Yeah, this looks good to me! Nice @xwvvvvwx !

PR #33, opened 2020-08-18
Simplifycases

MrChico:
based on top of #32 (which I am yet to review fully, but want to merge before).

Test vectors need updating, but this is the idea


src/Parse.y
```diff
@@ -197,7 +203,7 @@ Precondition : 'iff' nonempty(Expr)                   { Iff (pos $1) $2 }
 Store : Entry '=>' Expr                               { Rewrite $1 $3 }
       | Entry                                         { Constant $1 }
       
-Entry : id list(Zoom)                                 { Entry (pos $1) (arg $1) $2 }
+Entry : id list(Zoom)                                 { Entry (pos $1) (arg $1) $2 }
```
xwvvvvwx, line 206: no extraneous whitespace pls :pray: :sparkles:

src/Main.hs
```diff
@@ -189,33 +173,35 @@ splitBehaviour store (Transition name contract iface@(Interface _ decls) iffs' c
...
-      leaves <- mapM (flatten iff postc (c : pathcond)) (normalize cs)
-      return $ join leaves
+    normalize :: [Case] -> [Case]
+    normalize cases = snd $ mapAccumL checkCase (BoolLit False) cases
+
+    checkCase :: Expr -> Case -> (Expr, Case)
+    checkCase acc (Case p WildExp post) =
+      (error "wildcard not last case",
+        Case p (ENot nowhere acc) post)
+    checkCase acc (Case p e post) = (EOr nowhere e acc, Case p e post)
```
xwvvvvwx, line 188: I don't really understand what's going on here. Won't this fail for any list of cases that has a wildcard in it somewhere?

src/Main.hs
```diff
@@ -189,33 +173,35 @@ splitBehaviour store (Transition name contract iface@(Interface _ decls) iffs' c
...
+      (p, maybeReturn) <- checkPost env contract post
+      return $ splitCase name False contract iface (LitBool True) iff maybeReturn p postc
+    flatten iff postc (Branches branches) = do
+      let branches' = normalize branches
+      cases <- flip mapM branches $ \(Case pn cond post) -> do
+        if' <- checkBool env cond
+        (post', ret) <- checkPost env contract post
+        return (if', post', ret)
+
+      pure . join $ ((\(ifcond, ret, store) ->
```
xwvvvvwx, line 203: shouldn't this be `(ifcond, stateUpdates, ret)` or similar?

src/Main.hs
MrChico: My intention was to ensure that you can only have wildcard in the last case, as otherwise that case would be redundant. My thinking here is that we only refer to the first element of this tuple if we keep iterating through the array; otherwise we just consider the second element. But this is a bit of an ugly thing (not proper error handling), this is part of what makes this PR a draft, I suppose

src/Main.hs
MrChico: yes, I screwed up these argnames

src/K.hs
```diff
@@ -320,16 +320,18 @@ mkTerm this accounts Behaviour{..} invariant = (name, term)
...
                       pack $
                         kAccount pass a
                          (fromMaybe
                            (error $ show a ++ " not found in accounts: " ++ show accounts)
                            $ Map.lookup a accounts
                          )
-                         (fromMaybe [] $ Map.lookup a _stateUpdates)
-                    ) _contracts)))
+                         (fromMaybe
+                           (error $ "no stateUpdates for " ++ show a ++ " in " ++ show _stateUpdates)
```
MrChico, line 332: this might be a rebase fuckup. For pure functions there should be no stateupdates, I presume

PR #35, opened 2020-08-26
RFC: rounding errors

xwvvvvwx:
A little writeup of a proposed syntax for describing properties related to rounding error. 

[rendered](https://github.com/ethereum/act/blob/rounding-example/examples/rounding/readme.md)

### DIFF
--- a/examples/rounding/Pool.sol
+++ b/examples/rounding/Pool.sol
@@ -1,0 +1,90 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.6.12;
+
+interface ERC20 {
+    function balanceOf(address usr) external view returns (uint256);
+    function transfer(address usr, uint256 amt) external returns (bool);
+    function transferFrom(address src, address dst, uint256 amt) external returns (bool);
+}
+
+
+contract Pool {
+    // --- Data ---
+    string  public constant name = "Token";
+    string  public constant symbol = "TKN";
+    uint8   public constant decimals = 18;
+
+    ERC20 immutable public token0;
+    ERC20 immutable public token1;
+
+    uint256 public totalSupply;
+    mapping (address => uint) public balanceOf;
+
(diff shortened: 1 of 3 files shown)


MrChico:
Nice write up! I agree that this is a very nice thing to be able to express.

The annotations look a little ambiguous in their current location (how do we know that it refers to `amt1` here?).
An alternative would be to be a bit more explicit (and flexible) about what the `exact`ness is referring to, by doing explicit conversions to real numbers, maybe along the lines of something like this:

```
where
    amt0 := (token0.balanceOf[ACCT_ID] * shares) / totalSupply
    amt1 := (token1.balanceOf[ACCT_ID] * shares) / totalSupply

ensures
   amt0 <= toReal(token0.balanceOf[ACCT_ID]) * toReal(shares) / toReal(totalSupply)
   amt1 >= toReal(token1.balanceOf[ACCT_ID]) * toReal(shares) / toReal(totalSupply)
```

xwvvvvwx:
hmmmm yes, good point re: ambiguity. In my head the annotation refered to the whole expresison on the RHS of the storage rewrite, not just `amt1` or `amt0`. I think that I agree that being more explicit would be benficial.

I'm not sure about using `ensures`, would we be able to provide a `toReal` for the `K` and `hevm` backends? I think it might be confusing if only some of the claims in the postconditions are verified in those backends? We also lose the claims about the bounds of the error in your example.

I also think it would be harder to warn on expressions that do not have rounding annotations if we move them away from the storage writes and into the `ensures`.

Perhaps a dedicated `rounding` block at the behaviour level?

```
rounding

where
    amt0 := (token0.balanceOf[ACCT_ID] * shares) / totalSupply
    amt1 := (token1.balanceOf[ACCT_ID] * shares) / totalSupply

rounding
    amt0 <= exact(amt0), bound 1
    amt1 >= exact(amt1), bound 1
```

xwvvvvwx:
turned this into an issue at #112.

PR #37, opened 2020-08-27
Smt invariants

xwvvvvwx:
This is still pretty rough, need to fill in some gaps and write tests, but the code here is able to prove the invariant in `examples/homogeneous`, and displays a (very hard to read) counterexample if I change the invariant to e.g. `x * y = 2 * z`.

Opening as a draft for visibility.

## TODO

- [ ] tests
- [ ] pretty counterexamples
- [x] handle bytestrings
- [x] handle calldata args properly
- [x] handle pre/post conditions
- [x] handle external storage


src/Prove.hs
```diff
@@ -0,0 +1,249 @@
...
+data Ctx = Ctx Id Id When [(Id, SMType)]
+  deriving (Show)
+
+data SMType
+  = SymInteger (SBV Integer)
+  | SymBool (SBV Bool)
+  | SymBytes (SBV [WordN 8])
+  deriving (Show)
+
+catInts :: [(Id, SMType)] -> [(Id, SBV Integer)]
```
MrChico, line 55: another alternative would be to use `Data.Bifunctor` to do something along the lines of:
```
catInts = fmap $ second (\(SymInteger i) -> i)
```

src/Prove.hs
```diff
@@ -0,0 +1,249 @@
...
+queries claims = fmap mkQuery $ gather claims
+
+
+-- *** Data *** --
+
+
+data When = Pre | Post
+  deriving (Show)
+
+data Ctx = Ctx Id Id When [(Id, SMType)]
```
MrChico, line 46: can you annotate what the arguments are here? Haddock style could be cute

src/Prove.hs
```diff
@@ -0,0 +1,249 @@
...
+  -- TODO: refine AST so we don't need this anymore
+  when (contract /= c1 || contract /= c2 || c1 /= c2) $ error "Internal error: contract mismatch"
+
+  store <- mapM (makeSymbolic (Ctx contract method Pre [])) locs
+  let ctx = Ctx contract method Pre store
+
+  inv <- symExpBool ctx e                             -- invariant holds
+  state <- mapM (\(c, u) -> fromUpdate ctx u) (updates stateUpdates) -- state updates hold
+  -- TODO: handle constructor args
+  return $ (sAnd state) .&& (sNot inv)
```
MrChico, line 108: nit: these parens are not needed I think

src/Prove.hs
```diff
@@ -0,0 +1,249 @@
...
+  where
+    fromUpdate :: Ctx -> StorageUpdate -> Symbolic (SBV Bool)
```
MrChico, line 110: same as fromUpdate below? Can be factored out

src/Prove.hs
```diff
@@ -0,0 +1,249 @@
...
+nameFromItem :: Ctx -> TStorageItem a -> Id
+nameFromItem (Ctx contract method when _) item = case item of
+  DirectInt name -> contract @@ method @@ name @@ show when
+  DirectBool name -> contract @@ method @@ name @@ show when
+  DirectBytes name -> contract @@ method @@ name @@ show when
+  MappedInt name ixs -> contract @@ method @@ name @@ showIxs ixs @@ show when
+  MappedBool name ixs -> contract @@ method @@ name @@ showIxs ixs @@ show when
+  MappedBytes name ixs -> contract @@ method @@ name @@ showIxs ixs @@ show when
+  where
+    x @@ y = x <> "_" <> y
```
MrChico, line 234: nice

src/Prove.hs
```diff
@@ -0,0 +1,249 @@
...
+symExp ctx expression = case expression of
+  ExpInt e -> do
+    i <- symExpInt ctx e
+    return $ SymInteger i
+  ExpBool e -> do
+    b <- symExpBool ctx e
+    return $ SymBool b
+  ExpBytes e -> do
+    b <- symExpBytes ctx e
+    return $ SymBytes b
```
MrChico, line 183:
```suggestion
symExp :: Ctx -> ReturnExp -> Symbolic (SMType)
 symExp ctx expression = case expression of 
  
    ExpInt e -> SymInteger <$> symExpInt ctx e
 
   ExpBool e -> SymBool <$> symExpBool ctx e

   ExpBytes e -> SymBytes <$> symExpBytes ctx e
```

src/Prove.hs
```diff
@@ -0,0 +1,249 @@
...
+symExpInt :: Ctx -> Exp Integer -> Symbolic (SBV Integer)
+symExpInt ctx@(Ctx _ _ _ store) e = case e of
+  Add a b   -> (+)  <$> (symExpInt ctx a) <*> (symExpInt ctx b)
+  Sub a b   -> (-)  <$> (symExpInt ctx a) <*> (symExpInt ctx b)
+  Mul a b   -> (*)  <$> (symExpInt ctx a) <*> (symExpInt ctx b)
+  Div a b   -> sDiv <$> (symExpInt ctx a) <*> (symExpInt ctx b)
+  Mod a b   -> sMod <$> (symExpInt ctx a) <*> (symExpInt ctx b)
+  Exp a b   -> (.^) <$> (symExpInt ctx a) <*> (symExpInt ctx b)
+  LitInt a  -> return $ literal a
+  IntVar a  -> sInteger a -- TODO: handle calldata args properly
```
MrChico, line 214: It seems that if we keep calldata args in `Ctx` as well, this method (as well as `symExpBool`, `symExpBytes` and `symExp`) could be made pure and things would probably look a little cleaner.

src/Prove.hs
xwvvvvwx: Added some `newTypes` that I think document it in the type signature

src/Prove.hs
xwvvvvwx: hmmmm maybe, I think I'll leave them in cos it makes things a little more explicit.

src/Prove.hs
xwvvvvwx: slightly different unfortuantely as the one below operates on a pre and post state and this one just operates on a pre state.

src/Prove.hs
xwvvvvwx: you're nice :kissing:

src/Prove.hs
xwvvvvwx: good point! it's done.

src/Main.hs
```diff
@@ -92,14 +83,30 @@ main = do
...
                        Bad e -> prettyErr contents e
 
+      (Prove file solver smttimeout) -> do
+        contents <- readFile file
+        case parse (lexer contents) >>= typecheck of
+          Bad e -> prettyErr contents e
+          Ok claims -> do
+            let handleRes = \(SatResult res) -> case res of
+                              Unsatisfiable _ _ -> putStrLn "Q.E.D"
+                              Satisfiable _ model -> putStrLn $ "Counterexample found!\n" ++ show model
```
MrChico, line 93: we should exit with failure in failing cases

xwvvvvwx:
@MrChico I'm pretty happy with the core logic here now, gonna be focusing on adding tests over the next days. Feel free to review when you have some time :pray: :sparkles:

src/Main.hs
xwvvvvwx: done.

src/Main.hs
```diff
@@ -92,14 +83,47 @@ main = do
...
+                handleResults ((Invariant c e), rs) = do
+                  let msg = "\n============\n\nInvariant " <> show e <> " of " <> show c <> ": "
+                      sep = "\n\n---\n\n"
+                      results' = handleRes <$> rs
+                      ok = foldl (||) False $ fst <$> results'
+                  case ok of
+                    False -> putStrLn $ msg <> "Q.E.D ✨"
+                    True -> do
+                      putStrLn $ msg <> "\n\n" <> (intercalate sep $ snd <$> results')
+                      exitFailure
```
MrChico, line 101:
```suggestion
                  if ok of
                  then putStrLn $ msg <> "Q.E.D ✨"
                  else do
                      putStrLn $ msg <> "\n\n" <> (intercalate sep $ snd <$> results')
                      exitFailure
```

src/Main.hs
```diff
@@ -92,14 +83,47 @@ main = do
...
+        contents <- readFile file
+        case parse (lexer contents) >>= typecheck of
+          Bad e -> prettyErr contents e
+          Ok claims -> do
+            let
+                handleResults ((Invariant c e), rs) = do
+                  let msg = "\n============\n\nInvariant " <> show e <> " of " <> show c <> ": "
+                      sep = "\n\n---\n\n"
+                      results' = handleRes <$> rs
+                      ok = foldl (||) False $ fst <$> results'
```
MrChico, line 96:
```suggestion
                      ok = or $ fst <$> results'
```

src/Main.hs
```diff
@@ -92,14 +83,47 @@ main = do
...
+                  Satisfiable _ model -> (True, "Counter example found!\n\n" <> show model)
+                  Unknown _ reason -> (True, "Unknown! " <> show reason)
+                  ProofError _ reasons _  -> (True, "Proof error! " <> show reasons)
+                  SatExtField _ _ -> error "Extension field containing Infinite/epsilon"
+                  DeltaSat _ _ _ -> error "Unexpected DeltaSat"
+
+            results <- flip mapM (queries claims)
+                          (\(i, qs) -> do
+                            rs <- mapM (runSMTWithTimeOut solver smttimeout debug) qs
+                            pure $ (i, rs)
```
MrChico, line 114:
```suggestion
                            pure (i, rs)
```

src/Prove.hs
```diff
@@ -0,0 +1,340 @@
...
+   - constants for the pre and post versions of all storage variables used by the transition system
+   - boolean predicates over the pre and post storage variables for each pass behaviour
+   - a boolean predicate for the invariant
+   - an assertion that the invariant does not hold if either of the following is true:
+      1. The constructor predicate holds
+      2. The invariant holds over the prestate and one of the method level predicates holds
+
+   If this query returns `unsat` then the invariant must hold over the transition system
+-}
+queries :: [Claim] -> [(Invariant, [Symbolic ()])]
```
MrChico, line 37: what do you think of making this a Map rather than a list of pairs? It might be easier to read. And we could use the functor instance for map to map over its values like you are doing on https://github.com/ethereum/act/pull/37/files#diff-f8f3412da88cd4806f23d59fe59ebc3bR111.
It would add some `fromList` and `toList` instances here and there though

src/Prove.hs
```diff
@@ -0,0 +1,340 @@
...
+queries claims = fmap mkQueries gathered
+  where
+    gathered = fmap (\inv -> (inv, store, getBehaviours inv)) invariants
+    invariants = catInvs claims
+    store = head $ (catStores claims) -- TODO: refine AST so we don't need this head anymore
```
MrChico, line 42:
```suggestion
    store = head (catStores claims) -- TODO: refine AST so we don't need this head anymore
```

src/Prove.hs
```diff
@@ -0,0 +1,340 @@
...
+    getBehaviours (Invariant c _) = filter isPass $ filter (contractMatches c) (catBehvs claims)
```
MrChico, line 43:
```suggestion
    getBehaviours (Invariant c _) = filter (isPass .&& contractMatches c) (catBehvs claims)
```

src/Prove.hs
```diff
@@ -0,0 +1,340 @@
...
+
+  calldata <- Args <$> mapM (mkSymArg c m) decls
+  preStore <- Store <$> mapM (mkSymStorage m Pre) locs
+  postStore <- Store <$> mapM (mkSymStorage m Post) locs
+
+  let
+    preCtx = Ctx c m calldata preStore Pre
+    postCtx = Ctx c m calldata postStore Post
+
+  return $ (preCtx, postCtx)
```
MrChico, line 133:
```suggestion
  return (preCtx, postCtx)
```

src/Prove.hs
```diff
@@ -0,0 +1,340 @@
...
+    postCond' = symExpBool postCtx postCond
+    stateUpdates' = mkStorageConstraints preCtx postCtx stateUpdates locs
+
+  constrain $ preCond' .&& (sAnd stateUpdates') .&& postCond' .&& (sNot postInv')
+
+mkMethod :: Invariant -> Storages -> Behaviour -> Symbolic ()
+mkMethod inv@(Invariant _ e) (Storages store) behv@(Behaviour _ _ _ _ _ preCond postCond stateUpdates _) = do
+  (preCtx, postCtx) <- mkContexts inv behv
+
+  let
```
MrChico, line 102: style points

src/Prove.hs
```diff
@@ -0,0 +1,340 @@
...
+      where
+        lhs :: (Show b) => TStorageItem a -> ([(Id, SMType)] -> [(Id, b)]) -> b
+        lhs i f = get (nameFromItem m post i) (Map.fromList $ f postStore)
+
+
+-- *** Symbolic Expression Construction *** ---
+
+
+symExpBool :: Ctx -> Exp Bool -> SBV Bool
+symExpBool ctx@(Ctx c m (Args args) (Store store) w) e = case e of
```
MrChico, line 203: seems unused
```suggestion
symExpBool (Ctx c m (Args args) (Store store) w) e = case e of
```

src/Prove.hs
```diff
@@ -0,0 +1,340 @@
...
+-- *** Data *** --
+
+
+data When = Pre | Post
+  deriving (Eq, Show)
+
+newtype Contract = Contract { unContract :: Id }
+  deriving (Show)
+
+newtype Method = Method { unMethod :: Id }
```
MrChico, line 57: I'm not sure if I like these as newtypes or just simple type synonym:
```
type Method = Id
```

Then we won't need to bother with these constructors and stuff, it's really just readability

MrChico: Similar to how `Id` is already a synonym for `String`. What do you think?

src/Prove.hs
```diff
@@ -0,0 +1,340 @@
...
+
+references :: Invariant -> Behaviour -> [StorageLocation]
+references (Invariant _ inv) (Behaviour _ _ _ _ _ _ _ updates _)
+  = nub $ (getLoc <$> updates) <> locsFromExp inv
+
+mkSymArg :: Contract -> Method -> Decl -> Symbolic (Id, SMType)
+mkSymArg contract method decl@(Decl typ _) = case metaType typ of
+  Integer -> do
+    v <- sInteger name
+    return $ (name, SymInteger v)
```
MrChico, line 143:
```suggestion
    return (name, SymInteger v)
```
and sim. for other cases

src/RefinedAst.hs
```diff
@@ -63,71 +69,78 @@ data MType
...
+  DirectInt    :: Id -> Id -> TStorageItem Integer
+  DirectBool   :: Id -> Id -> TStorageItem Bool
+  DirectBytes  :: Id -> Id -> TStorageItem ByteString
+  MappedInt    :: Id -> Id -> NonEmpty ReturnExp -> TStorageItem Integer
+  MappedBool   :: Id -> Id -> NonEmpty ReturnExp -> TStorageItem Bool
+  MappedBytes  :: Id -> Id -> NonEmpty ReturnExp -> TStorageItem ByteString
 
 deriving instance Show (TStorageItem a)
+
+instance Eq (TStorageItem a) where
```
MrChico, line 94: can this be derived?
```
deriving instance Eq (TStorageItem a)
```

src/Type.hs
```diff
@@ -111,64 +106,82 @@ splitBehaviour store (Transition name contract iface@(Interface _ decls) iffs' c
...
-
--- extract a list of iff headers from the size of the types in a list of calldata declarations
+    mkBound :: Either StorageLocation StorageUpdate -> Maybe (Exp Bool)
+    mkBound ref = case ref of
+      Left loc -> case loc of
+        IntLoc item -> Just $ fromItem item
+        _ -> Nothing
+      Right update -> case update of
+        IntUpdate item _ -> Just $ fromItem item
+        _ -> Nothing
```
MrChico, line 167:
```suggestion
    mkBound (Left loc) = case loc of
        IntLoc item -> Just $ fromItem item
        _ -> Nothing
     mkBound (Right update) = case update of
        IntUpdate item _ -> Just $ fromItem item
        _ -> Nothing
```

src/Type.hs
```diff
@@ -111,64 +106,82 @@ splitBehaviour store (Transition name contract iface@(Interface _ decls) iffs' c
...
+
+    fromItem :: TStorageItem Integer -> Exp Bool
+    fromItem item = case item of
+      DirectInt contract name -> bound (abiType $ slotType contract name) (TEntry item)
+      MappedInt contract name _ -> bound (abiType $ slotType contract name) (TEntry item)
```
MrChico, line 172:
```suggestion
    fromItem (DirectInt contract name) = bound (abiType $ slotType contract name) (TEntry item)
    fromItem (MappedInt contract name _) = bound (abiType $ slotType contract name) (TEntry item)
```

Makefile
```diff
@@ -21,13 +21,15 @@ compiler: bin/act
 
 test_specs=$(wildcard tests/*/*.act)
 
-failing_typing=tests/array/array.act tests/dss/vat.act tests/creation/createMultiple.act
-
-typing_specs=$(filter-out $(failing_typing), $(test_specs))
+parser_specs=$(filter-out $(invariant_specs), $(test_specs))
```
MrChico, line 24: why not test the parsing of invariants?

src/Prove.hs
xwvvvvwx: This needs an `Ord` instnace for `Invariant`, and by extension `Exp Bool`, which the compiler is not able to derive by itself, I'm not sure if it's worth the boilerplate?

What do you think?

src/Prove.hs
xwvvvvwx: The nice thing is that they ensure you can't e.g. accidentally pass an `Args` instead of a `Store`, so we gain a little typeee safety with them, but I think I agree that they are overkill here....

src/RefinedAst.hs
xwvvvvwx: oh, yes it can, dunno how I missed that. Thanks!!!

Makefile
xwvvvvwx: hmmmm, I just didn't want to clutter up the tests folder with a bunch of parsed output...

xwvvvvwx: I'm not sure how valuable I find the parsing tests atm anyway, the output is extememly hard to diff and read. I would be more in favour of removing them and fuzzing the parser instead...

src/Prove.hs
MrChico: Right, ok, let's leave it as is

src/Prove.hs
MrChico: Yeah, I think it's a little overkill

MrChico: imho, this is more about "types as documentation" then giving any extra safety props

Makefile
MrChico: Ok, that's alright with me

MrChico (approved):

PR #39, opened 2020-09-24
Constant invariants

xwvvvvwx:
This allows proving of invariants like [`totalSupply == _totalSupply`](https://github.com/ethereum/act/compare/constant-invariants?diff=unified&expand=1#diff-7161029c5459643e61ddebdb0119e1c2R13) in `transfer.act`.

The approach in this PR is to add the arguments from the constructor to the arguments in the method context.

Opening as a draft for now because this approach is probably unsound in cases where there are arguments to methods with the same name as constructor args. @MrChico do you have any ideas on how to handle this more cleanly?


src/Prove.hs
```diff
@@ -119,9 +128,14 @@ mkContexts inv@(Invariant contract _) behv@(Behaviour method _ _ c1 (Interface _
   postStore <- Map.fromList <$> mapM (mkSymStorage method Post) locs
   env <- mkEnv contract method
 
+  calldata' <- case initBehv of
+                 Nothing -> pure $ calldata
+                 Just (Behaviour _ _ _ _ (Interface _ initDecls) _ _ _ _) -> do
+                   initCd <- Map.fromList <$> mapM (mkSymArg contract method) initDecls
+                   pure $ Map.unionWith (\l _ -> l) initCd calldata
```
MrChico, line 135:
```suggestion
                   pure $ Map.keys $ initCd <> calldata
```

MrChico, line 135: ummm, I think this is just normal union. Though I guess you want something like `Map.unionWith (error "TODO: duplicated calldata arguments")` for now or something

xwvvvvwx: oh didn't know about normal union haha :grinning:

xwvvvvwx:
@MrChico implemented using a seperate map for the constructor args as we discussed in chat. 

I ended up needing to have two contexts again though, but I think this is the cleanest way...

MrChico:
For some reason it really bothers me to create two contexts from mkContexts.
What about if we leave mkContexts as it is, and have mkMethod deconstruct it like:
```
ctx@(Ctx c m arg store env) <- mkContexts ...
```
and create `invCtx` directly in mkMethod as:
```
args <- mkArgs initdecls
let invCtx = Ctx c m args store env
```

xwvvvvwx:
@MrChico hehe, tbh I think they are both a little awkward, but I'm fine with either way.

Implemented as suggested above, take a look :slightly_smiling_face:

MrChico:
yes I much prefer this!

PR #43, opened 2020-09-25
Bound eth env

xwvvvvwx:
Travereses the AST for any `EthEnv` type expressiona and if found adds bounds for them as `if` style preconditions to all behaviours.

The full diff is a little cluttered due to some extraneous tidyups, the last commit contains the important logic.

### DIFF
--- a/src/Type.hs
+++ b/src/Type.hs
@@ -12,6 +12,7 @@ import EVM.ABI
 import EVM.Solidity (SlotType(..))
 import Data.Map.Strict    (Map)
 import Data.Maybe
+import Data.Either
 import Data.List.NonEmpty (NonEmpty(..))
 import qualified Data.List.NonEmpty as NonEmpty
 import qualified Data.Map.Strict    as Map -- abandon in favor of [(a,b)]?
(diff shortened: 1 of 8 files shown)


src/Type.hs
```diff
@@ -135,22 +140,70 @@ splitBehaviour store (Constructor name contract iface@(Interface _ decls) iffs (
...
 splitCase name creates contract iface if' iffs ret storage postcs =
-  [ B $ Behaviour name Pass creates contract iface (mconcat (if':iffs)) (mconcat postcs) storage ret,
-    B $ Behaviour name Fail creates contract iface (And if' (Neg (mconcat iffs))) (mconcat postcs) storage Nothing ]
+  [ B $ Behaviour name Pass creates contract iface (mconcat (if' <> iffs)) (mconcat postcs) storage ret,
+    B $ Behaviour name Fail creates contract iface (And (mconcat if') (Neg (mconcat iffs))) (mconcat postcs) storage Nothing ]
+
+mkEthEnvBounds :: [EthEnv] -> [Exp Bool]
+mkEthEnvBounds vars = catMaybes $ mkBound <$> nub vars
+  where
+    mkBound :: EthEnv -> Maybe (Exp Bool)
```
MrChico, line 167: why not do this in terms of `bound` instead?

xwvvvvwx: much cleaner, even helped me find a little bug :pray: :sparkles:

MrChico (approved):

PR #44, opened 2020-09-26
coq backend

endorphin:


src/RefinedAst.hs
```diff
@@ -117,6 +117,11 @@ data Exp t where
   LitInt :: Integer -> Exp Integer
   IntVar :: Id -> Exp Integer
   IntEnv :: EthEnv -> Exp Integer
+  -- bounds
```
endorphin, line 120: it's helpful for the coq backend to abstract the integer bounds

src/RefinedAst.hs
```diff
@@ -117,6 +117,11 @@ data Exp t where
...
+  IntMin :: Int -> Exp Integer
+  IntMax :: Int -> Exp Integer
+  UIntMin :: Int -> Exp Integer
+  UIntMax :: Int -> Exp Integer
```
endorphin, line 124: it's helpful for the coq backend to abstract the integer bounds

src/K.hs
```diff
@@ -146,6 +146,10 @@ kExprInt (Div a b) = "(" <> kExprInt a <> " /Int " <> kExprInt b <> ")"
 kExprInt (Mod a b) = "(" <> kExprInt a <> " modInt " <> kExprInt b <> ")"
 kExprInt (Exp a b) = "(" <> kExprInt a <> " ^Int " <> kExprInt b <> ")"
 kExprInt (LitInt a) = show a
+kExprInt (IntMin a) = kExprInt $ LitInt $ 0 - 2 ^ (a - 1)
+kExprInt (IntMax a) = kExprInt $ LitInt $ 2 ^ (a - 1) - 1
+kExprInt (UIntMin a) = kExprInt $ LitInt 0
+kExprInt (UIntMax a) = kExprInt $ LitInt $ 2 ^ a - 1
```
endorphin, line 152: I think this is all that's needed for K, but I don't really understand the SMT implementation. @MrChico @xwvvvvwx can you help?

src/RefinedAst.hs
endorphin: @MrChico @xwvvvvwx can you review the related fixes in https://github.com/ethereum/act/pull/44/commits/2da2812d3ee84fe8518a3f94be713fd28bcbb45b and https://github.com/ethereum/act/pull/44/commits/5122289a6e5ff60f58ea2ce1859863f30435d5a5?

Makefile
```diff
@@ -12,7 +12,7 @@ src/Lex.hs: src/Lex.x
 
 # builds the rest of the haskell files (compiler)
 bin/act: src/*.hs
-	cd src && cabal new-update && cabal v2-install --installdir=../bin --overwrite-policy=always && cd ..
```
xwvvvvwx, line 15: The `new-update` is an unfortunate workaround to a confusing purity leak in the `nix-shell`, CI currently fails without it. I think it's nice to remove it from the `Makefile` because it's very annoying to wait for the update, but maybe lets add it to the github workflow instead?

default.nix
```diff
@@ -28,6 +28,9 @@ let
       pkgs.jq
     ];
     withHoogle = true;
+    shellHook = ''
+      export PATH=$PWD/bin:$PATH
```
xwvvvvwx, line 32:
```suggestion
      export PATH=${toString ./bin}:$PATH
```

default.nix
```diff
@@ -28,6 +28,9 @@ let
       pkgs.jq
```
xwvvvvwx, line 28:
```suggestion
      pkgs.jq
      pkgs.coq
```

src/Coq.hs
```diff
@@ -0,0 +1,285 @@
...
+stateval store handler updates =
+
+  "state " <> T.intercalate " "
+    (map (valuefor (rights updates)) (M.toList (headval store)))
+
+  where
+
+  valuefor :: [StorageUpdate] -> (Id, SlotType) -> T.Text
+  valuefor updates' (name, t) =
+    case listToMaybe (filter (f name) updates') of
```
xwvvvvwx, line 139: You will need to import `Data.List` for this one:

```suggestion
    case find (f name) updates' of
```

src/Coq.hs
```diff
@@ -0,0 +1,285 @@
...
+-- | coq syntax for a return expression
+retexp :: ReturnExp -> T.Text
+retexp (ExpInt e) = coqexp e
+retexp (ExpBool e) = coqexp e
+retexp (ExpBytes _) = error "bytestrings not supported"
+
+-- | coq syntax for a list of arguments
+coqargs :: NE.NonEmpty ReturnExp -> T.Text
+coqargs (e NE.:| es) =
+  retexp e <> " " <> T.intercalate " " (map retexp es) where
```
xwvvvvwx, line 278:
```suggestion
  retexp e <> " " <> T.intercalate " " (map retexp es)
```

src/Prove.hs
```diff
@@ -415,3 +427,8 @@ catBytes m = Map.fromList $ go $ Map.toList m
     go ((name, SymBytes b):tl) = (name, b):(go tl)
     go (_:tl) = go tl
     go [] = []
+
+intmin a = 0 - 2 ^ (a - 1)
```
xwvvvvwx, line 431:
```suggestion
intmin a = negate $ 2 ^ (a - 1)
```

xwvvvvwx, line 431: please add type signatures

xwvvvvwx (commented):
very nice surprise to wake up to!!!! :sparkles: :smile_cat: 

Just a few nits for now...

Makefile
endorphin: https://github.com/ethereum/act/pull/44/commits/1d765b1402e7e21b0fdd8efba17a155eb6f70166 should be sufficient, right?

endorphin:
fixes are in https://github.com/ethereum/act/pull/44/commits/12898b74c570eb1042e7264dd0fa35162132fac5

Makefile
xwvvvvwx: yes :pray: :sparkles:

examples/safemath/Theory.v
```diff
@@ -93,5 +96,4 @@ Proof.
     unfold not. intros. discriminate.
 
   }
-Qed.
-
+Qed. Check mul_correct.
```
xwvvvvwx, line 99: What does the `Check` do?

xwvvvvwx:
Do we need to reason about words and modulo arithmetic here? The proofs in the smt backend operate on integers only. We can use the smt solver to check that all expressions do not exceed the bounds of their type, and spec authors can add explicit cases if overflow is expected.

What do you think @MrChico ?

src/RefinedAst.hs
MrChico: those changes look good to me

default.nix
MrChico: we should consider making coq an optional dependency

MrChico:
@xwvvvvwx so far, the idea has been to let Act specs only refer to integers, and similarly all of the operations defined are interpreted as integer operations. The word semantics defined [here](https://github.com/ethereum/act/pull/44/files#diff-2ab23355dd0886c440d2bcb1961ddf98R289-R305) does not follow this pattern, and I think it would be cleaner if the coq backend could avoid word arithmetic completely.

I think it's a good idea to check that expressions cannot exceed their bound in the smt backend as you say @xwvvvvwx. Ideally with an error message that says smth like "Did you mean to add `iff in range uin256 x + y` to the spec?".
That the actual implementation actually respects these overflow bounds would be a proof obligation that would be dispatched to HEVM / KEVM / the SMT-Checker.

MrChico:
This is super cool and much further along than I expected! I tried some basic experiments but could not get the generated coq libs to interface correctly yet, but I suspect that is a "user problem".
It would be nice if the definitions of invariants was generated as well, with a hole put in place of proof obligations

examples/safemath/Theory.v
endorphin: It's just a vernacular command to typecheck an expression. I added it to print the statement of the theorem to stdout when building, but feel free to remove it.

examples/transitions/state_machine.act
```diff
@@ -1,39 +1,32 @@
...
+    x <= 2
 
 behaviour f of StateMachine
 interface f()
 
 case x == 0:
-	storage
-		x => 1
-
-case _:
```
MrChico, line 16: what does this machine do when `x /= 0`? This is currently unspecified. Should it be `iff x == 0` instead?

MrChico: If we are to follow the [code](https://github.com/ethereum/act/blob/coq/examples/transitions/state_machine.sol), we should probably keep the `case _: noop`  specified

src/Coq.hs
```diff
@@ -0,0 +1,410 @@
...
+  <> block (evalSeq (base store)  <$> groups constructors)
+  <> reachable (groups constructors) (groups transitions)
+
+  where
+
+  behaviours = [a | B a <- claims]
+
+  transitions = filter f behaviours where
+    f (Behaviour _ Pass False _ _ _ _ _ _) = True
+    f _ = False
```
MrChico, line 58:
```suggestion
  transitions = filter (not . _creation) behaviours
```

src/Coq.hs
```diff
@@ -0,0 +1,410 @@
...
+
+  constructors = filter f behaviours where
+    f (Behaviour _ Pass True  _ _ _ _ _ _) = True
+    f _ = False
```
MrChico, line 62:
```suggestion
  constructors = filter _creation behaviours
```

src/Coq.hs
```diff
@@ -0,0 +1,410 @@
...
+
+-- | multiline implication
+implication :: [T.Text] -> T.Text
+implication xs = "   " <> T.intercalate "\n-> " xs
+
+-- | wrap text in parentheses
+parens :: T.Text -> T.Text
+parens s = "(" <> s <> ")"
+
+mod256 :: T.Text -> T.Text
```
MrChico, line 365: is this still used?

src/Coq.hs
```diff
@@ -0,0 +1,410 @@
...
+import EVM.Solidity (SlotType(..))
+import Syntax
+import RefinedAst
+
+type Store = M.Map Id SlotType
+type Fresh = State Int
+
+header :: T.Text
+header =
+  "(* --- GENERATED BY ACT --- *)\n\n\
```
MrChico, line 34: here (and in general throughout the code) it might be nice to use `unlines` to make things more readable

src/Coq.hs
MrChico: or if we introduce lenses for `Behaviour`:
`filter (not . view creation)`

src/Coq.hs
```diff
@@ -0,0 +1,410 @@
...
+  reachableType "" (stateType <> " -> " <> stateType <> " -> Prop") body
+
+  where
+
+  body = concat $
+    (evalSeq baseCase <$> constructors)
+    <>
+    (evalSeq reachableStep <$> groups)
+
+  reachableStep :: Behaviour -> Fresh T.Text
```
MrChico, line 89: In general, I prefer promoting functions to toplevel when they don't rely on locally scoped variables. This is the case for `reachableStep` and `baseCase`, so I think they should be promoted.

src/Coq.hs
```diff
@@ -0,0 +1,410 @@
...
+        <> universal <> "\n"
+        <> constructorBody where
+      baseval = parens $ name' <> " " <> arguments i
+      constructorBody = (indent 2) . implication . concat $
+        [ coqprop <$> conds
+        , [reachableType <> " " <> baseval <> " " <> baseval]
+        ]
+      universal = case length decls of
+        0 -> ""
+        _ -> "forall " <> interface i <> ","
```
MrChico, line 121:
```suggestion
      universal = if null decls
        then ""
        else "forall " <> interface i <> ","
```

src/Coq.hs
```diff
@@ -0,0 +1,410 @@
...
+  f n (IntUpdate (MappedInt _ n' _) _)
+    | n == n' = True
+  f n (BoolUpdate (DirectBool _ n') _)
+    | n == n' = True
+  f n (BoolUpdate (MappedBool _ n' _) _)
+    | n == n' = True
+  f _ _ = False
+
+  -- represent mapping update with anonymous function
+  lambda :: [ReturnExp] -> Int -> Exp a -> Id -> T.Text
```
MrChico, line 191: promote to top level

src/Coq.hs
```diff
@@ -0,0 +1,410 @@
...
+returnType (ExpInt _) = "Z"
+returnType (ExpBool _) = "bool"
+returnType (ExpBytes _) = "bytestrings not supported"
+
+-- | default value for a given type
+-- this is used in cases where a value is not set in the constructor
+defaultValue :: SlotType -> T.Text
+defaultValue t =
+
+  case t of
```
MrChico, line 240: I would do this pattern matching directly in the function LHS

src/Coq.hs
```diff
@@ -0,0 +1,410 @@
...
+abiType AbiStringType = strMod <> ".string"
+abiType a = error $ show a
+
+-- | coq syntax for a return type
+returnType :: ReturnExp -> T.Text
+returnType (ExpInt _) = "Z"
+returnType (ExpBool _) = "bool"
+returnType (ExpBytes _) = "bytestrings not supported"
+
+-- | default value for a given type
```
MrChico, line 235: actually, all state variables must be defined and given an initial value in the constructor right now, so this seems superfluous

src/K.hs
```diff
@@ -353,6 +357,6 @@ mkTerm this accounts Behaviour{..} invariant = (name, term)
                   )
                <> "\nrequires "
                <> defaultConditions (kVar _contract) <> "\n andBool\n"
-               <> kExprBool _preconditions
+               <> kExprBool (foldr (\a b -> And a b) (LitBool True) _preconditions)
```
MrChico, line 360:
```suggestion
               <> kExprBool (mconcat _preconditions)
```

src/Main.hs
```diff
@@ -113,6 +117,18 @@ main = do
                           )
             mapM_ handleResults results
 
+      (Coq f) -> do
+        contents <- readFile f
+        case parse (lexer contents) of
```
MrChico, line 122: I made a continuation passing style `proceed` function so we can avoid all of these nested cases

src/Prove.hs
```diff
@@ -87,17 +87,17 @@ mkQueries (inv, store, behvs) = (inv, inits <> methods)
...
 mkInit :: Invariant -> Behaviour -> Symbolic ()
-mkInit inv@(Invariant _ e) behv@(Behaviour _ _ _ _ _ preCond postCond stateUpdates _) = do
+mkInit inv@(Invariant _ e) behv@(Behaviour _ _ _ _ _ preConds postCond stateUpdates _) = do
   ctx <- mkContext inv behv
 
   let
     mkBool = symExpBool ctx
     postInv' = mkBool Post e
-    preCond' = mkBool Pre preCond
+    preConds' = mkBool Pre (foldr (\a b -> And a b) (LitBool True) preConds)
```
MrChico, line 96:
```suggestion
    preConds' = mkBool Pre (mconcat preConds)
```

src/Prove.hs
```diff
@@ -111,12 +111,13 @@ mkMethod inv@(Invariant _ e) (Storages rawStorage) initBehv behv = do
     locs = references inv behv
     preInv = symExpBool invCtx Pre e
     postInv = symExpBool invCtx Post e
-    preCond = symExpBool ctx Pre (_preconditions behv)
+    preConds = symExpBool ctx Pre $
+      foldr (\a b -> And a b) (LitBool True) (_preconditions behv)
```
MrChico, line 115:
```suggestion
      mconcat (_preconditions behv)
```

MrChico:
Very nice, this looks quite clean!
The `Exponent.v` file seems to be missing from the example generated.
It would still be nice to:
- Be able to generate the Makefile, and the whole coq project to a designated directory `act coq --file exponent.v --out ./examples/exponent`.
- Generate proof obligations from invariants. It seems like this wouldn't be too far away since `coqprop` is already written.

But these things don't need to go into this PR. I think this is good to merge when my comments are addressed.

examples/transitions/state_machine.act
endorphin: you're right, thanks for pointing that out. i'm not sure why i removed the catch-all case. patched in https://github.com/ethereum/act/commit/59aa11c9eb54ba1f41ef9b89b9b2508124d6c967. personally, i think it's cleaner to use an `iff`.

src/Coq.hs
endorphin: patched in https://github.com/ethereum/act/commit/48b6c628a093f6ab32fb5173ca5996e6f999fabe

src/Coq.hs
endorphin: nope, removed

src/Coq.hs
endorphin: patched in https://github.com/ethereum/act/commit/c5c6e3e7ce127f2a7a165985c8165d00e69f23c6. coq isn't newline sensitive in a lot of cases, so I only used it in places where it makes sense to emphasize the line structure.

src/Coq.hs
endorphin: done (https://github.com/ethereum/act/commit/dd1cacdc86e6f5dc80cda1b611fb3e8b183b7638)

src/Coq.hs
endorphin: done (https://github.com/ethereum/act/commit/d7f7027327d6b4682445886dc960fa1b0e6f0cc7).

src/Coq.hs
endorphin: it's not superfluous because of mappings. `mapping(t => t') m := []` doesn't have a sensible meaning if there isn't a default value for the type `t'`.

src/Main.hs
endorphin: fixed in https://github.com/ethereum/act/pull/44/commits/1c2f3209d39d211b8337bbf713e13fefb0ae8652

examples/transitions/state_machine.act
MrChico: hmm, but now it doesn't match the code. I'm fine with either, but if you want `iff` you need to change the code to use `requires` instead of `if`

src/Coq.hs
MrChico: Right, ok, this is good

examples/transitions/state_machine.act
endorphin: oops, right

endorphin:
> The `Exponent.v` file seems to be missing from the example generated.

I didn't check it in because it's a generated file. The other examples don't include the generated coq models either.

MrChico:
> > The `Exponent.v` file seems to be missing from the example generated.
> 
> I didn't check it in because it's a generated file. The other examples don't include the generated coq models either.

ok fair. But perhaps it would be good to make a test out of these files, which I guess basically amounts to just invoking the Makefiles in the subdirectories, so we can ensure these things will continue to work in the ci

Makefile
```diff
@@ -30,10 +30,18 @@ invariant_fail=$(wildcard tests/invariants/fail/*.act)
 
 failing_typing=tests/array/array.act tests/dss/vat.act tests/creation/createMultiple.act
 
+coq-examples = examples/transitions examples/safemath examples/exponent
+
+.PHONY: test-coq $(coq-examples)
+test-coq: compiler $(coq-examples)
+$(coq-examples):
+	make -C $@
```
endorphin, line 38: how's this @MrChico ?

MrChico: amazin

MrChico (approved):

PR #45, opened 2020-09-29
Lex: enforce british spelling of behaviour

xwvvvvwx:


leonardoalt:
`examples/transitions/state_machine.act` uses `behavior`, also `tools/act.py`. I can adjust the script later, but it might make sense to update the example in this PR as well.

xwvvvvwx:
oups missed those. thanks @leonardoalt!

leonardoalt (approved):

Issue #38, opened 2020-09-10
Incorrect AST output for `token` specs


xwvvvvwx:
There should be some storage updates in the first `stateUpdates` block [here](https://github.com/ethereum/act/blob/b91cb38438cccd713eb985a6222bee1c83a502f4/tests/token/transfer.act.typed.json), but there are only `StorageLocation`'s...

xwvvvvwx:
the output in the json was simply out of order due to https://github.com/ethereum/act/pull/47...

PR #47, opened 2020-09-29
Parserev

MrChico:
Our current parsing setup reversed all lists...
@xwvvvvwx I think this might also solve #38 ?


xwvvvvwx:
> @xwvvvvwx I think this might also solve #38 ?

Yes it looks that why, but I don't really understand why?

MrChico:
I think things might have just been out of place there

xwvvvvwx:
wdym out of place?

MrChico:
I think things might have been correct, we were just looking in the wrong place because everything was reversed. So the "first case" was actually the last

xwvvvvwx:
hmmmm no I don't think that was the case, if you look here: https://github.com/ethereum/act/blob/b91cb38438cccd713eb985a6222bee1c83a502f4/tests/token/transfer.act.typed.json#L3 the `stateUpdates` block is for the `transferFrom` `Pass` spec and contains only `Left` values.

MrChico:
yes, but the last case of `transferFrom` also contains no state updates:
```
case src == dst:

     balanceOf[CALLER]
     allowance[src][CALLER]
     balanceOf[src]
     balanceOf[dst]

```

xwvvvvwx:
ohhhhhhhhhhh yes.

PR #48, opened 2020-09-30
Fix ethEnv in SMT backend

xwvvvvwx:
Had made a small oversight and forgotten to remove some "TODO" errors... :unamused: 

Additionally renamed `Address` to `This` in the AST.

Thanks for catching this @leonardoalt !


tests/invariants/fail/ethEnv.act
```diff
@@ -0,0 +1,48 @@
...
+    uint timestamp := TIMESTAMP
+    uint nonce := NONCE
+
+invariants
+
+    this == THIS
+    value == CALLVALUE
+    depth == CALLDEPTH
+    // hash == BLOCKHASH
+    number == BLOCKNUMBER
```
xwvvvvwx, line 25: Needs polymorphic Eq...

tests/invariants/fail/ethEnv.act
```diff
@@ -0,0 +1,48 @@
...
+    value => 1
+    depth => 1
+    number => 1
+    difficulty => 1
+    chainid => 1
+    gaslimit => 1
+    coinbase => 1
+    timestamp => 1
+    nonce => 1
+    // hash => (hash ++ BLOCKHASH)
```
xwvvvvwx, line 48: Needs support for `ECat` in `inferExpr`

tests/invariants/pass/ethEnv.act
```diff
@@ -0,0 +1,31 @@
...
+    uint difficulty := DIFFICULTY
+    uint chainid := CHAINID
+    uint gaslimit := GASLIMIT
+    address coinbase := COINBASE
+    uint timestamp := TIMESTAMP
+    uint nonce := NONCE
+
+invariants
+
+    this == THIS
```
MrChico, line 21: hmm, so the semantics here is that any env var appearing in the invariant is interpreted as the value of that env var when the contract was created. I think that makes sense

leonardoalt: @MrChico why is that? I'd say that should be true for `THIS` only

xwvvvvwx: hmmm true this spec is actually pretty weird. The actual smt query will create a fresh constant for each env var for every method, and then the invariant will be expressed in terms of that constant. So I don't think the claims in smt make any connection to the value of the env var when the contract was created...

leonardoalt: > The actual smt query will create a fresh constant for each env var for every method

I think that makes sense, except for `THIS` which in my opinion should always be the same variable, as long as we're talking about the same contract.

leonardoalt: Actually `CHAINID` as well

xwvvvvwx: hmmmmm yes that does make sense, but it's probably going to be annoying to implement. The queries for each method are totally independant at the moment... 

I also think it's a little  unclear to mix semantics for env vars...

MrChico: One reason why I think it should be the initial value is because the `invariant` block lives under the same scoping as the constructor, and for the constructor, any env var is obviously set to the value at constructor time. So if we let envvars change over time we would have two different interpretations there. 
The counterpoint to this argument is that this is already true for storage vars.

leonardoalt: >  The counterpoint to this argument is that this is already true for storage vars.

Exactly, the interpretation for state variables is already "the value of the state variable at transaction T", where T is universally quantified.

xwvvvvwx: Agree that both interpretations seem useful. Perhaps we should avoid ambiguity by requiring that each reference to the environment must be tagged with either as either `initial` or `variable`?

We could also disallow some of the weird edge cases this way  (e.g. `variable(THIS)`).

Issue #42, opened 2020-09-25
Fail specs should not contain stateUpdates


xwvvvvwx:
fail specs currently contain state updates, but they should instead leave the state exactly as it was before. e.g: https://github.com/ethereum/act/blob/52476d8db2bd2301d767e3920b6475a6cf31e902/tests/token/transfer.act.typed.json#L1628-L1758

PR #51, opened 2020-10-10
Frontend fuzzing

xwvvvvwx:
Apologies for the size. There are quite a few small tweaks and unrelated changes, but the main changes here are:

1. Add a pretty printer `Print.hs`
2. Split the current typechecking pass into two phases, one that type checks (`Type.hs`), and one that adds conditions based on the types of the variables (`Enrich.hs`)
3. Use QuickCheck and the pretty printer to fuzz the lexer / parser / typechecker by generating a random behaviour, printing it, running it back through the lex / parse/ typecheck pipeline, and then asserting that the result is the same as the generated behaviour. This intended to surface cases where the refined AST and the act spec are non isomorphic.

Some smaller changes:

- Add position information to typecheck errors
- Squash compiler warnings in `Main.hs` (shadowed variables)
- Fix a bug in `Type.hs` where `Fail` specs were incorrectly being given nonEmpty `returns` and `postconditions` values as the `Pass` specs (a similar bug remains unfixed for `storage` blocks: https://github.com/ethereum/act/issues/42).
- Added some missing JSON serialization code
- Makes `postconditions` an `[Exp Bool]` instead of an `Exp Bool`
- Represent numbers as strings in the JSON output (as suggested by @MrChico in previous PR)

The QuickCheck generator is currently missing support for:

1. storage entries / updates (more complex to ensure that the generated behaviours typecheck)
2. ITE for `Exp ByteString` and `Exp Bool` (currently unsupported in the frontend)
3. ByteString literals, as well as the cat and slice operators (unsupported in the frontend)
4. `ByStr`: @MrChico It's not clear to me what this is actually supposed to represent? Is it a literal or a variable?
5. Negative integer literals (unsupported in the frontend)
6. `case` blocks (these do not exist at the level of the refined AST)

Note that splitting the typecheck pass also fixed a bug in the typechecker where the preconditions derived from the storage and calldata bounds were added `iff` style preconditions, and so were negated for the fail specs. `enrich` currently adds preconditions derived from the type of `EthEnv`, storage and calldata vars to all `Pass` and `Fail` specs, which I believe to be sound.


xwvvvvwx:
@endorphin I'm not sure what's going on with the `Coq` tests? I get:

```
Error: The reference STATE was not found in the current environment.
```

but I didn't touch the Coq backend at all, except in `Main.hs` to split `typecheck` into `enrich <$> typecheck`. Could I have messed something up with my changes in `Type.hs`?

endorphin:
Looks like the enrich phase is adding storage preconditions to constructors (behaviours with the `_creation` flag).

```
behaviour init of StateMachine
interface constructor()

creates
  uint256 x := 0

invariants
    x <= 2
```
Previously, the `Pass` behaviour corresponding to `init` had an empty list of preconditions. After this set of changes, the preconditions are:
```
[And (LEQ (LitInt 0) (TEntry (DirectInt "StateMachine" "x"))) (LEQ (TEntry (DirectInt "StateMachine" "x")) (UIntMax 256))]
```
The coq backend assumes that constructors don't have preconditions involving storage. There shouldn't be any storage to talk about before a constructor is called.

xwvvvvwx:
@endorphin Thanks! That fixed it. I'm still not sure why I was getting the same failure on master, I suspect I had some old `.v` files lying around after switching branches...

src/Enrich.hs
```diff
@@ -0,0 +1,158 @@
...
+import qualified Data.Map.Strict as Map (lookup)
+
+import EVM.ABI (AbiType(..))
+import EVM.Solidity (SlotType(..))
+
+import RefinedAst
+import Type (bound, defaultStore, metaType)
+import Syntax (EthEnv(..), Id, Decl(..), Interface(..))
+
+-- |Adds extra preconditions to non constructor behaviours based on the types of their variables
```
MrChico, line 16:
```suggestion
-- | Adds extra preconditions to non constructor behaviours based on the types of their variables
```

src/Enrich.hs
```diff
@@ -0,0 +1,158 @@
...
+      pre' = pre
+             <> (mkCallDataBounds decls)
+             <> (mkStorageBounds store stateUpdates)
+             <> (mkEthEnvBounds $ ethEnvFromBehaviour behv)
+
+ethEnvFromBehaviour :: Behaviour -> [EthEnv]
+ethEnvFromBehaviour (Behaviour _ _ _ _ _ preconds postconds stateUpdates returns) =
+  (concat $ ethEnvFromExp <$> preconds)
+  <> (concat $ ethEnvFromExp <$> postconds)
+  <> (concat $ ethEnvFromStateUpdate <$> stateUpdates)
```
MrChico, line 42:
```suggestion
  (concatMap ethEnvFromExp preconds)
  <> (concatMap ethEnvFromExp postconds)
  <> (concatMap ethEnvFromStateUpdate stateUpdates)
```

src/Enrich.hs
```diff
@@ -0,0 +1,158 @@
...
+  Left (BytesLoc item) -> ethEnvFromItem item
+  Right (IntUpdate item e) -> ethEnvFromItem item <> ethEnvFromExp e
+  Right (BoolUpdate item e) -> ethEnvFromItem item <> ethEnvFromExp e
+  Right (BytesUpdate item e) -> ethEnvFromItem item <> ethEnvFromExp e
+
+ethEnvFromItem :: TStorageItem a -> [EthEnv]
+ethEnvFromItem item = case item of
+  MappedInt _ _ ixs -> concat $ ethEnvFromReturnExp <$> ixs
+  MappedBool _ _ ixs -> concat $ ethEnvFromReturnExp <$> ixs
+  MappedBytes _ _ ixs -> concat $ ethEnvFromReturnExp <$> ixs
```
MrChico, line 58:
```suggestion
  MappedInt _ _ ixs -> concatMap ethEnvFromReturnExp ixs
  MappedBool _ _ ixs -> concatMap ethEnvFromReturnExp ixs
  MappedBytes _ _ ixs -> concatMap ethEnvFromReturnExp ixs
```

src/Main.hs
```diff
@@ -108,55 +107,58 @@ main = do
...
-      setEnv "SBV_CVC4_OPTIONS" ""
-      return res
-  | solver == Just "z3" = runwithz3
-  | solver == Nothing = runwithz3
-  | otherwise = error "Unknown solver. Currently supported solvers; z3, cvc4"
- where timeout = fromMaybe 20000 maybeTimeout
-       runwithz3 = satWith z3{verbose=debug'} $ (setTimeOut timeout) >> sym
+satWithTimeOut solver' maybeTimeout debug' sym = case solver' of
+  Just "cvc4" -> do
+    setEnv "SBV_CVC4_OPTIONS" ("--lang=smt --no-interactive-prompt --model-witness-value --tlimit-per=" <> show timeout)
```
MrChico, line 144: hmm, why did you remove `--incremental` and `--interactive`/

MrChico:
I think my idea with `ByStr` was to admit two different ways of generating Bytes, either through a hex string as in: `0xdead` or from a string `"Hello"`.

src/Main.hs
xwvvvvwx: oh, I had thought they were only useful in the context of `hevm`, where we are using the `query` monad and `push`/`pop`?

So far in `act` we just submit queries to the solver fully formed and call `check-sat` once per query. Did I misunderstand something?

MrChico: We might be able to get away without them, but keeping them doesn't seem very problematic either. I put them all there because those were [sbvs default options](https://github.com/LeventErkok/sbv/blob/master/Data/SBV/Provers/CVC4.hs#L31)

xwvvvvwx: ah ok, I'll put them back.

xwvvvvwx:
> I think my idea with ByStr was to admit two different ways of generating Bytes, either through a hex string as in: 0xdead or from a string "Hello". So ByStr would be literal strings.

Ah ok that makes sense. I guess that means they are currently printed wrong [here](https://github.com/ethereum/act/pull/51/files#diff-51cf038532963fb6a740b663a5eaa6ecR77), but since we don't seem to have support for them in the frontend I guess it's OK to leave them like that for now?

MrChico:
The printing looks okay for now.
This looks good to me! The fuzz tests are being run on ci and all of the changes are good

MrChico (approved):

PR #52, opened 2020-10-12
Restructure tests dir a bit

leonardoalt:
- Created `tests/frontend` for parsing/typing tests
- `invariants` stays unmodified
- Created `tests/coq` and moved the `coq` tests there. In the future `hevm` will also use those files, so not sure if it's better to change later or just come up with something now. Also some tests are duplicated or almost duplicated, for example SafeMath. It exists in both `frontend` and `coq`, with slightly different versions.


MrChico:
This looks good to me, but I could also see the SafeMath examples being unified right here in this PR as well

MrChico (approved):

leonardoalt:
> This looks good to me, but I could also see the SafeMath examples being unified right here in this PR as well

@MrChico Yea exactly. The question then is, do we remove it from `frontend`, do we unify but duplicate it in `frontend` and `coq`?

Issue #29, opened 2020-07-09
[Act 0.1] Document the language


leonardoalt:
We should go through the structure of the language and each section

xwvvvvwx:
I think for `v0.1` the following should be sufficient:

1. A simple language reference (similar to [acts.md](https://github.com/dapphub/klab/blob/master/acts.md)).
2. A nice blogpost similar to the [hevm symbolic](https://fv.ethereum.org/2020/07/28/symbolic-hevm-release/), or [ds-test](https://fv.ethereum.org/2020/12/11/symbolic-execution-with-ds-test/).

For later releases I would really like to have a more comprehensive documentation in a gitbook or similar.

PR #57, opened 2020-11-10
act syntax description

MrChico:
This aims to be a start at addressing #29

### DIFF
--- a/docs/act.md
+++ b/docs/act.md
@@ -39,7 +39,7 @@ A constructor section consists of the following fields:
 Specifying the arguments the constructor takes.
 Example:
 ```act
-interface constructor(uint _owner)
+interface constructor(address _owner)
 ```
 
 ### iff (optional)
@@ -109,7 +109,7 @@ invariants
 A `behaviour` specification is indicated by the
 `behaviour of <contractName>` header, as in:
 ```act
-constructor of A
+behaviour of A
 ```
 
 and consists of the following fields:
--- a/docs/act.md
+++ b/docs/act.md
@@ -25,7 +25,7 @@ definitions of [RefinedAst.hs](https://github.com/ethereum/act/blob/master/src/R
 
 ## Types
 
-The types of Act are made up from three basic primitives:
+The types of Act consist of three basic primitives:
 Integers, Booleans and ByteStrings. Integers are unbounded,
 with true integer operations. However, as our integer expressions
 will often represent words in the EVM, we allow ourselves a slight
--- a/docs/act.md
+++ b/docs/act.md
@@ -32,7 +32,7 @@ will often represent words in the EVM, we allow ourselves a slight
 abuse of notation and denote by `uintN` integers together with the
 constraint that the value fits into a `uintN`.
 
-Using conventional abi types for typing also allows us to specify
+Using conventional ABI types for typing also allows us to specify
 function signatures in a concise way.
 
 As an example consider the specification of overflow safe addition:
--- a/docs/act.md
+++ b/docs/act.md
@@ -29,8 +29,8 @@ The types of Act consist of three basic primitives:
 Integers, Booleans and ByteStrings. Integers are unbounded,
 with true integer operations. However, as our integer expressions
 will often represent words in the EVM, we allow ourselves a slight
-abuse of notation and denote by `uintN` integers together with the
-constraint that the value fits into a `uintN`.
+abuse of notation and denote by `uintN/intN` integers together with the
+constraint that the value fits into a `uintN/intN`.
 
 Using conventional ABI types for typing also allows us to specify
 function signatures in a concise way.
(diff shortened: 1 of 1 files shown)


docs/act.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+A constructor section consists of the following fields:
+
+### interface
+
+Specifying the arguments the constructor takes.
+Example:
+```act
+interface constructor(uint _owner)
```
xwvvvvwx, line 42:
```suggestion
interface constructor(address _owner)
```

docs/act.md
```diff
@@ -0,0 +1,195 @@
...
+### creates
+Defines the storage layout of the contract.
+
+Example:
+```act
+
+creates
+
+  uint256 totalSupply := _totalSupply
+  mapping(address => uint) balanceOf :=  [CALLER := _totalSupply]
```
xwvvvvwx, line 68: would be nice to have a little info on the allowed types here, as well as the initializer list syntax.

docs/act.md
```diff
@@ -0,0 +1,195 @@
...
+
+and consists of the following fields:
+
+
+### interface
+
+Specifying which method the behaviour refers to.
+Example:
+```act
+interface transfer(address to, uint value)
```
xwvvvvwx, line 123: we should note the allowed types here

docs/act.md
```diff
@@ -0,0 +1,195 @@
...
+
+Example:
+```act
+iff
+
+  balanceOf[CALLER] >= value
+  CALLVALUE == 0
+```
+
+### cases and state updates
```
xwvvvvwx, line 140: Maybe worth adding a note that either `returns` or `storage` are required, but not both?

xwvvvvwx (approved):
nice!!!

Can we add a link to here in the readme somewhere?

docs/act.md
```diff
@@ -11,18 +11,51 @@ Each `behaviour` specification determines how a contract method updates
...
 
-This document is a high level description of the syntax of 
+This document is a high level description of the syntax of
 act specifications.
 For a more formal treatment of the syntax, study the
 definitions of [RefinedAst.hs](https://github.com/ethereum/act/blob/master/src/RefinedAst.hs).
 
+## Types
+
+The types of Act are made up from three basic primitives:
```
leonardoalt, line 28:
```suggestion
The types of Act consist of three basic primitives:
```

docs/act.md
```diff
@@ -11,18 +11,51 @@ Each `behaviour` specification determines how a contract method updates
...
+Integers, Booleans and ByteStrings. Integers are unbounded,
+with true integer operations. However, as our integer expressions
+will often represent words in the EVM, we allow ourselves a slight
+abuse of notation and denote by `uintN` integers together with the
+constraint that the value fits into a `uintN`.
```
leonardoalt, line 33: Is this also valid for `int`?

docs/act.md
```diff
@@ -11,18 +11,51 @@ Each `behaviour` specification determines how a contract method updates
...
+
+Using conventional abi types for typing also allows us to specify
```
leonardoalt, line 35:
```suggestion
Using conventional ABI types for typing also allows us to specify
```

docs/act.md
```diff
@@ -11,18 +11,51 @@ Each `behaviour` specification determines how a contract method updates
...
+
+   x + y
+
+returns x + y
+```
+
+In more verbose terms, this specification would read:
+
+Given any pair of integers x and y, s.t. 0 <= x < 2^256 and
+0 <= y < 2^256, an abiencoded call to the contract `SafeMath`
```
leonardoalt, line 54:
```suggestion
0 <= y < 2^256, an ABI encoded call to the contract `SafeMath`
```

docs/act.md
MrChico: yes. The semantics then is that `x : intN` is `x : int s.t. 2 ^ (N - 1) <= x < 2 ^ (N -1 )`

MrChico: will add to clarify

docs/act.md
```diff
@@ -0,0 +1,230 @@
...
+definitions of [RefinedAst.hs](https://github.com/ethereum/act/blob/master/src/RefinedAst.hs).
+
+## Types
+
+The types of Act consist of three basic primitives:
+Integers, Booleans and ByteStrings. Integers are unbounded,
+with true integer operations. However, as our integer expressions
+will often represent words in the EVM, we allow ourselves a slight
+abuse of notation and denote by `uintN` integers together with the
+constraint that the value fits into a `uintN`.
```
MrChico, line 33:
```suggestion
will often represent words in the EVM, we allow ourselves a slight
abuse of notation and denote by `uintN/intN` integers together with the
constraint that the value fits into a `uintN/intN`.
```

PR #68, opened 2020-12-15
Add positions to (almost) all type error messages

xwvvvvwx:
- Extends position information to all syntax elements
- Presents that information to users for (almost) all type error messages
- Removes the redundent `ETrue` and `EFalse` syntax elements

I couldn't figure out how to add positions to the following errors:

- [multiple contract definitions](https://github.com/ethereum/act/blob/b78ffd6715323dd93ee36aaf9d6035963e3e80f0/src/Type.hs#L41)
- [cannot assign multiple values to an atomic type](https://github.com/ethereum/act/blob/b78ffd6715323dd93ee36aaf9d6035963e3e80f0/src/Type.hs#L168)


src/Syntax.hs
```diff
@@ -97,18 +93,18 @@ data Expr
 --    | Look Pn Id [Expr]
     | Func Pn Id [Expr]
     | ListConst Expr
-    | EmptyList
+    | EmptyList Pn
```
xwvvvvwx, line 96: This is unused atm. Do we need it @MrChico?

MrChico: lets move it right now

src/Lex.x
```diff
@@ -211,9 +213,16 @@ showposn (AlexPn _ line column) =
 posn :: Lexeme -> AlexPosn
 posn (L _ p) = p
 
-arg :: Lexeme -> String
-arg (L (ID s) _) = s
-arg _ = error "TODO: support all lexemes in arg"
+arg :: Lexeme -> LEX
```
MrChico, line 216: Hmm, I think I'd prefer to just add
```
value :: Lexeme -> Integer
value (L (ILIT i) _) = i
```
and keep `arg` as it is. This way there would be a lot less functions applied in Parse.y

MrChico (approved):

PR #69, opened 2020-12-15
Add new tests

leonardoalt:


tests/invariants/fail/type_constraints_env.act
```diff
@@ -0,0 +1,8 @@
+constructor of AllEnvVars
+interface constructor()
+
+creates
+	uint value := CALLVALUE
```
leonardoalt, line 5: This test should pass. Added to `fail` so that this passes for now, and fails when it's fixed.

tests/invariants/fail/constructor_indirect_assignment.act
```diff
@@ -0,0 +1,11 @@
...
+interface constructor()
+
+creates
+
+	uint y := x
+	uint x := 2
+
+invariants
+
+	y == 2
```
leonardoalt, line 11: This test should pass. Added to fail so that this passes for now, and fails when it's fixed.

Issue #66, opened 2020-12-15
SMT backend env vars type constraints


leonardoalt:
```
constructor of AllEnvVars
interface constructor()

creates
	uint value := CALLVALUE

invariants
	value >= 0
```

I would expect that to be always true based on value's type, but:
```
$ ./result/bin/act prove --file tests/invariants/pass/all_env_vars.act 

============

Invariant "(AllEnvVars.value >= 0)" of "AllEnvVars": 

Counter example found!

SMTModel {modelObjectives = [], modelBindings = Nothing, modelAssocs = [("AllEnvVars_init_value",0 :: Integer),("AllEnvVars_init_value_post",-1 :: Integer),("AllEnvVars_init_CALLER",0 :: Integer),("AllEnvVars_init_CALLVALUE",-1 :: Integer),("AllEnvVars_init_CALLDEPTH",0 :: Integer),("AllEnvVars_init_ORIGIN",0 :: Integer),("AllEnvVars_init_BLOCKHASH","" :: String),("AllEnvVars_init_BLOCKNUMBER",0 :: Integer),("AllEnvVars_init_DIFFICULTY",0 :: Integer),("AllEnvVars_init_CHAINID",0 :: Integer),("AllEnvVars_init_GASLIMIT",0 :: Integer),("AllEnvVars_init_COINBASE",0 :: Integer),("AllEnvVars_init_TIMESTAMP",0 :: Integer),("AllEnvVars_init_THIS",0 :: Integer),("AllEnvVars_init_NONCE",0 :: Integer)], modelUIFuns = []}
```

leonardoalt:
```
constructor of AllEnvVars
interface constructor()

creates
	uint value := 5

invariants
	value >= 0
```
This holds, so I guess it's the env vars that are missing the type constraints.

Issue #60, opened 2020-12-15
Polymorphic Equality


xwvvvvwx:
The equality operator is currently only defined over integers ([ref](https://github.com/ethereum/act/blob/master/src/RefinedAst.hs#L96)). We should allow `Eq` over at least booleans as well.

PR #84, opened 2020-12-18
Constructor bounds

xwvvvvwx:
Fixes https://github.com/ethereum/act/issues/66.

Adds preconditions to all constructors ensuring that:

1. All local storage vars are initially zero
1. All environment variables are within the range of their type
1. All external storage is within the range of it's type
1. All calldata variables are within the range of their type


xwvvvvwx:
[`6bdc0dd`](https://github.com/ethereum/act/pull/84/commits/622694fb8d13b67c1e4b635d7d6c0f79cbae1577) adds a prestate to the constructor definitions in the Coq backend. While at first glance this might seem a little weird (there isn't a local prestate when the constructor is running), this simplifies the code in the SMT backend, and more importantly prepares us for handling constructors that reference or write to external storage in the Coq backend (where a prestate does exist in the external contracts). 

This PR already constrains integer values to zero in the constructor precondiditons, and I will extend that to force bytestrings to be empty and booleans to be false.

cc @endorphin @MrChico does this make sense?

xwvvvvwx:
Added support for polymorphic `Eq` in the AST. This allows comparison of `Bool` and `Bytes` types in act specs. This closes #60.

src/K.hs
```diff
@@ -157,16 +162,21 @@ kExprBool :: Exp Bool -> String
...
 kExprBool (LitBool a) = show a
 kExprBool (BoolVar a) = kVar a
+kExprBool (NEq a b) = "notBool (" <> kExprBool (Eq a b) <> ")"
+kExprBool (Eq (a :: Exp t) (b :: Exp t)) = case eqT @t @Integer of
+  Just Refl -> "(" <> kExprInt a <> " ==Int " <> kExprInt b <> ")"
+  Nothing -> case eqT @t @Bool of
+    Just Refl -> "(" <> kExprBool a <> " ==Bool " <> kExprBool b <> ")"
+    Nothing -> case eqT @t @ByteString of
+      Just Refl -> "(" <> kExprBytes a <> " ==K " <> kExprBytes b <> ")" -- TODO: Is ==K correct?
+      Nothing -> error "Internal Error: invalid expression type"
```
xwvvvvwx, line 179: @MrChico this is a little clunky, is there some nice trick with the Maybe monad I can use to tidy this up a bit?

MrChico: hmm, the fact that you are proceeding in the `Nothing` case makes it a bit tricky... I don't immediately see a cleaner way

xwvvvvwx:
Made some small tweaks to the `K` backend:

- rm the logic that added the invariant to the postconditions (this was incomplete and broken as implemented)
- add indentation in the generated K output
- add support for a few more kinds of storage location

xwvvvvwx:
Reverted the Coq changes after looking over them together with @MrChico. I will remove the local constructor prestate from the SMT backend in a seperate PR to avoid having this one get any more messy than it already is.

Summary of the changes made:

- Adds new preconditions to constructor specs:
  - All environment variables are within the range of their type
  - All external storage is within the range of it's type
  - All calldata variables are within the range of their type
  
- Adds support for polymorphic Eq expressions

- Tweaks to the K backend:
  - rm the logic that added the invariant to the postconditions (this was incomplete and broken as implemented)
  - add indentation in the generated K output
  - add support for a few more kinds of storage location

MrChico (approved):

MrChico:
Sweet! Looks good!

endorphin:
just got to this. looks good to me, and I guess this doesn't need more review since it's already merged and the coq changes were reverted.

PR #90, opened 2021-01-14
Enfore build cleanliness

xwvvvvwx:
This PR adds a `ci` flag to the build system that:

- forces all warnings to be errors
- enables the optimizer (`-O2`)

The `ci` flag is enabled for all `nix` builds.

Local builds (`cabal v2-build`) will still succeed in the prescense of warnings (although the same set of warnings will be shown in both configs).

We currently ignore the following compiler warnings in both configurations:

- orphan instances
- ~deprecated~

This is built on top of #89, I'll retarget to master once it's merged.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -12,9 +12,10 @@ let
       hevm = dapptools.haskellPackages.hevm;
       sbv = dapptools.haskellPackages.sbv;
       act =
-        hself.callCabal2nix
+        hself.callCabal2nixWithOptions
           "act"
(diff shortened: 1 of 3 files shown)


MrChico (approved):

MrChico:
what's the procedure if I happen to dislike one of the tips?

xwvvvvwx:
> what's the procedure if I happen to dislike one of the tips?

Just add it as an ignored warning.

PR #101, opened 2021-05-03
Recursion schemes 2

kjekac:
Ok, prepare for a monster. Sorry. :(

This PR introduces the usage of recursion schemes, using the package [compdata](https://hackage.haskell.org/package/compdata-0.12.1). With some small changes to our AST type, the recursion schemes from that package allow us to manipulate/create/collapse ASTs in a more principled – and ideally/sometimes also less verbose – way. In principle, *every function* which deals recursively with an AST (whether it be creation, mutation or destruction), can be implemented using a recursion scheme.

Unfortunately, most cases seem to be more or less as verbose as before, so I'm questioning whether it's actually worth it or not. The disadvantage is that recursion schemes aren't trivial to grok. Not very hard, but not trivial. The advantage is, again, that the manipulation of ASTs can be done in a more principled way, and given that you already understand recursion schemes, the manipulation is actually conceptually simpler in a certain way. So the tradeoff is essentially: code that is harder to get into but easier to work with, or the reverse?

Below, I try to explain this as clearly as possible without getting too caught up in details, so that we can decide which path we want to take. I'll start by showing a few examples of recursion schemes in our code base without any in-depth explanation, and then talk about them a bit more generally.

As a reference, here is a list of all the functions which I've refactored using recursion schemes so far. Some of these are duplicates to show different approaches. Most of these will be discussed below, but not all.

- [`locsFromExp`](https://github.com/ethereum/act/blob/recursion-schemes-2/src/Extract.hs#L40) uses `cataK`
- [`locsFromExp'`](https://github.com/ethereum/act/blob/recursion-schemes-2/src/Extract.hs#L47) uses `cata`
- [`ethEnvFromExp`](https://github.com/ethereum/act/blob/recursion-schemes-2/src/Extract.hs#L91) uses `cataK`
- [`kExpr`](https://github.com/ethereum/act/blob/recursion-schemes-2/src/K.hs#L120) uses `paraK`
- [`coqexp`](https://github.com/ethereum/act/blob/recursion-schemes-2/src/Coq.hs#L305) uses `cataK`
- [`coqprop`](https://github.com/ethereum/act/blob/recursion-schemes-2/src/Coq.hs#L309) uses `zygoK`
- [`coqprop'`](https://github.com/ethereum/act/blob/recursion-schemes-2/src/Coq.hs#L325) uses `zygo`

### Replacing repetitive cases

Let's start with a very nice example of when recursion schemes work well for us. In `Extract.hs`, we can replace this:

```haskell
locsFromExp :: Exp a -> [StorageLocation]
locsFromExp e = case e of
  And a b   -> (locsFromExp a) <> (locsFromExp b)
  Or a b    -> (locsFromExp a) <> (locsFromExp b)
  Impl a b  -> (locsFromExp a) <> (locsFromExp b)
  Eq a b    -> (locsFromExp a) <> (locsFromExp b)
  LE a b    -> (locsFromExp a) <> (locsFromExp b)
  LEQ a b   -> (locsFromExp a) <> (locsFromExp b)
  GE a b    -> (locsFromExp a) <> (locsFromExp b)
  GEQ a b   -> (locsFromExp a) <> (locsFromExp b)
  NEq a b   -> (locsFromExp a) <> (locsFromExp b)
  Neg a     -> (locsFromExp a)
  Add a b   -> (locsFromExp a) <> (locsFromExp b)
  Sub a b   -> (locsFromExp a) <> (locsFromExp b)
  Mul a b   -> (locsFromExp a) <> (locsFromExp b)
  Div a b   -> (locsFromExp a) <> (locsFromExp b)
  Mod a b   -> (locsFromExp a) <> (locsFromExp b)
  Exp a b   -> (locsFromExp a) <> (locsFromExp b)
  Cat a b   -> (locsFromExp a) <> (locsFromExp b)
  Slice a b c -> (locsFromExp a) <> (locsFromExp b) <> (locsFromExp c)
  ByVar _ -> []
  ByStr _ -> []
  ByLit _ -> []
  LitInt _  -> []
  IntMin _  -> []
  IntMax _  -> []
  UIntMin _ -> []
  UIntMax _ -> []
  IntVar _  -> []
  LitBool _ -> []
  BoolVar _ -> []
  NewAddr a b -> locsFromExp a <> locsFromExp b
  IntEnv _ -> []
  ByEnv _ -> []
  ITE x y z -> locsFromExp x <> locsFromExp y <> locsFromExp z
  TEntry a -> storageLocations a
```

with this:

```haskell
locsFromExp :: Exp a -> [StorageLocation]
locsFromExp = cataK \case
  BoolStore t -> storageLocations t
  IntStore t  -> storageLocations t
  ByStore t   -> storageLocations t
  e           -> hfold e
```

In case you're not familiar with the `\case` syntax, it is simply an anonymous function which allows pattern matching on the input. It is available via the GHC language extension `LambdaCase`. You are probably wondering what `cataK` and `hfold` do and why we've replaced `TEntry` with `BoolStore`, `IntStore` and `ByStore`, but hopefully the general idea is clear: we're able to substitute a single call to `hfold` for all cases which simply concatenate the results from a straightforward recursion. Much nicer.

We'll get to the specifics of `cataK` and `hfold` and why we replaced `TEntry` later, but for now let's move on to a different example.

### Factoring out the recursion

In `Coq.hs` we have `coqexp`, shown here with many cases omitted to reduce repetition:

```haskell
coqexp :: Exp a -> T.Text

-- booleans
coqexp (LitBool True)  = "true"
coqexp (BoolVar name) = T.pack name
coqexp (And e1 e2)  = parens $ "andb "   <> coqexp e1 <> " " <> coqexp e2
coqexp (Eq e1 e2)   = parens $ coqexp e1  <> " =? " <> coqexp e2
coqexp (Neg e)      = parens $ "negb " <> coqexp e
coqexp (TEntry (DirectBool _ name)) = parens $ T.pack name <> " " <> stateVar
coqexp (TEntry (MappedBool _ name args)) = parens $
  T.pack name <> " s " <> coqargs args

-- integers
coqexp (LitInt i) = T.pack $ show i
coqexp (Add e1 e2) = parens $ coqexp e1 <> " + " <> coqexp e2
coqexp (IntMin n)  = parens $ "INT_MIN "  <> T.pack (show n)


-- polymorphic
coqexp (ITE b e1 e2) = parens $ "if "
  <> coqexp b
  <> " then "
  <> coqexp e1
  <> " else "
  <> coqexp e2
```

This function doesn't lend itself to code reduction in the same way as `locsFromExp` does, because every case needs to be treated in a unique way. So at first glance, the version which uses a recursion scheme actually looks *more* complicated (the same cases as before are omitted):

```haskell
-- | coq syntax for an expression
coqexp :: Exp a -> T.Text
coqexp = cataK expAlg

expAlg = \case

--booleans
  LitBool b    -> T.toLower . T.pack . show $ b
  BoolVar name -> T.pack $ name
  And e1 e2  -> prefix2 "andb"  <$*> e1 <$*> e2
  Eq e1 e2   -> infix2  "=?"    <$*> e1 <$*> e2
  Neg e      -> prefix1 "negb"  <$*> e
  BoolStore (DirectBool _ name) -> prefix1 (T.pack name) stateVar
  BoolStore (MappedBool _ name args) -> prefix2 (T.pack name) "s" (coqargs args)

-- integers
  LitInt i    -> T.pack . show $ i
  Add e1 e2 -> infix2  "+"        <$*> e1 <$*> e2
  IntMin n  -> prefix1 "INT_MIN"  (T.pack $ show n)

-- polymorphic
  ITE b e1 e2 -> (\cond true false -> apply ["if",cond,"then",true,"else",false])
                  <$*> b <$*> e1 <$*> e2

apply :: [T.Text] -> T.Text
apply = parens . T.intercalate " "

prefix1 :: T.Text -> T.Text -> T.Text
prefix1 op x   = apply [op,x]

prefix2, infix2 :: T.Text -> T.Text -> T.Text -> T.Text
prefix2 op x y = apply [op,x,y]
infix2  op x y = apply [x,op,y]
```

Except for that weird operator `<$*>`, this seems to be more or less exactly the same function as before (with the text concatenation factored out into the helper functions at the bottom), right? *Wrong!*

Look again. As weird as `<$*>` may be (spoiler: it's *extremely simple* and actually isomorphic to `$`), it doesn't know that we're working in the context of `coqexp`, so it *cannot* make any recursive calls back to `coqexp`. And `coqexp` doesn't make any recursive calls either. So where exactly does the recursion happen?

The answer is in the recursion scheme `cataK`. What recursion schemes do is that they *factor out the recursion* into a higher-order function, so that we can write functions that are completely non-recursive and focus on more on the *what* and less on the *how*.

To drive home this point, what do you think the type of `e1` in the case for `And` is? Its name may be misleading, because it is *not* an expression. In fact, we have `e1 :: K Text Bool`, which is isomorphic to `Text`. As we'll see in a bit, we have to use the constant functor `K` to keep the type system happy, but conceptually you can always think of `K a b` as simply `a`, with no loss of precision.

So the arguments inside the constructors are actually the *results of this function applied to that argument*. We only have to think about how to recurse *once*: when choosing which recursion scheme to use. Much less room for mistakes.

### More complicated recursions

Continuing in `Coq.hs`, let's have a look at `coqprop` (with omitted cases):

```haskell
coqprop :: Exp a -> T.Text
coqprop (LitBool True)  = "True"
coqprop (And e1 e2)  = parens $ coqprop e1 <> " /\\ " <> coqprop e2
coqprop (Eq e1 e2)   = parens $ coqexp e1 <> " = "  <> coqexp e2
```

This is actually a place where I initially misunderstood the code. I thought that all recursive cases were simply calling `coqprop`, but it turns out that a few of them call `coqexp` from our previous example. Using the recursion scheme `zygoK` allows us to make this more explicit:

```
coqprop :: Exp Bool -> T.Text
coqprop = zygoK expAlg \case
 LitBool b              -> T.pack . show $ b
 And  (Snd e1) (Snd e2) -> infix2  "/\\" e1 e2
 Eq   (Fst e1) (Fst e2) -> infix2  "="   e1 e2
```

First, note that in contrast to `cataK`, `zygoK` takes *two* reduction functions – or *algebras* – as arguments. The first is `expAlg`, which we used to defined `coqexp` in the previous example. The second is a new algebra in which each pattern matching case explicitly states which algebra should be used to produce the result – `Fst` for the first and `Snd` for the second. (If we wanted, we could even get both by matching on `Both a b`!)

## Dispelling the magic

Now that you've seen a few examples of we can use recursion schemes, let's look behind the scenes. First off, as you may have guessed by now, `cataK` and `zygoK` aren't "real" recursion schemes. They're helper functions from `Utils.hs` which I've defined around `cata` and `zygo`, to avoid explicitly having to deal with getting the types around `K` to match up over and over again. For example, look at the literal definition of `cataK`:

```haskell
cataK f = unK . (cata $ K . f)
```

If we keep in mind that `K a b` is isomorphic to `a` and simply ignore the bookkeeping that the type system requires us to do, this is literally just `cataK f = cata f`.

With that in mind, there are a few things we need to explain at this point:

1. Why have we made changes to the AST type? (Replacing `TEntry`.)
1. What is `K` and why do we use it?
1. What is `h`, as in `hfold`?
1. What are `cata` and `zygo`?

Let's take these in order:

### 1. The "flat" expression type `ExpF`

If you have a look in `RefinedAst.hs`, this PR not only replaces `TEntry`, it actually changes the *kind* of our AST type. We used to have this:

```haskell
data Exp t where
  And  :: Exp Bool -> Exp Bool -> Exp Bool
  Or   :: Exp Bool -> Exp Bool -> Exp Bool
  ...
```

But now we have this:

```haskell
data ExpF r t where
  And  :: r Bool -> r Bool -> ExpF r Bool
  Or   :: r Bool -> r Bool -> ExpF r Bool
  ...

type Exp = Term ExpF
```

Reiterating some points from before, the whole idea behind recursion schemes is that we can *factor out the recursion*. When defining functions, we do this using the higher-order functions `cata`, `zygo` among others. But we also need to make the corresponding change to the data type. We factor out the recursion into the type parameter `r :: * -> *`. Try replacing `r` with our old type `Exp`, or maybe even `ExpF r`, and you'll see the idea!

But `r` can be replaced with many other things! When implementing `coqexp` earlier, we had `r ~ K Text`. In that case, the recursive component had been replaced with the result of a computation. By factoring out the recursive component, we allow recursion schemes to replace it with the result of running the computation on that component.

This allows us to implement functions on a recursive data type by dealing with a *corresponding flat datatype*. So we only ever have to care about exactly one "layer" of the data type, we never have to recurse manually, and leave less room for error!

Now, what is a `Term`? As you can see above, in addition to changing `Exp` into `ExpF`, we also define a type synonym:

```haskell
type Exp = Term ExpF
```

So it allows us to recover a fully recursive type from the "flattened" type. If we wanted to do this manually, we need to parametrize an `ExpF` on an `ExpF` on an `ExpF` on an `ExpF` ad infinitum, like this:

```haskell
Exp t ≅ ExpF (ExpF (ExpF (ExpF (...)))) t
```

But this is impossible to type, so instead we can use `Term ExpF t`. That is, conceptually speaking, `Term` is just a fixpoint. And if we need to unroll one layer of the recursion, we can use the function `unTerm :: Term f t -> f (Term f) t`, which with `f ~ ExpF` becomes `unTerm :: Term ExpF t -> ExpF (Term ExpF) t`.

Exactly how `Term` is implemented is kind of complicated and not terribly important for our purposes, but if you're curious you can have a look at [the documentation](https://hackage.haskell.org/package/compdata-0.12.1/docs/Data-Comp-Multi-Term.html#t:Term). 

Now, *finally*, the reason why we need to replace `TEntry` is kind of stupid:

We need some notion of equality on `ExpF`, but the types of the terms that will be passed to the constructors are unknown (they are all `r Bool` and similar), so this is difficult. The `compdata` package provides some type classes for higher-ordered heterogenous equality which work around this. Because the equality is *heterogenous*, we can't assume that the two ASTs being compared are parametrized by the same type. Hence, the `TStorageItem i` inside the first `TEntry` may not have the same `i` as that inside the second. So we cannot pass them to `==` as we previously did. Introducing constructors like `BoolStore :: TStorageItem Bool -> ExpF r Bool` solves this, because then the constructors tell us whether we can, and indeed need to, apply `==`.

### 2. Briefly on the constant functor `K`

The easiest way to understand `K` is to look at its [definition](https://hackage.haskell.org/package/compdata-0.12.1/docs/Data-Comp-Multi-HFunctor.html#t:K):

```haskell
newtype K a i = K {unK :: a} deriving (Functor, Foldable, Traversable)
```

In short: `i` is the type parameter which we can `fmap` over, but it is also a *phantom type*. No term ever has the type `i`. Instead, the only purpose of `K` is to embed the type `a` in a functor.

For example, if we have `K "foo" :: K String Int`, we can pass it to `fmap (>5) :: Functor f => f Int -> f Bool` and get back `K "foo" :: K String Bool`. The embedded `"foo" :: String` is completely untouched.

If we actually wanted to do something with the `String` inside, we could pattern match it out, or use `unK :: K a b -> b`. But when implementing `coqexp`, we already saw another way:

```haskell
(<$*>) :: (a -> b) -> K a x -> b
f <$*> (K a) = f a
```

This allows us to apply a function to the *actual* contents of the functor, instead of operating on the phantom type. Similarly, we have functions corresponding to `fmap` and `<*>`:

```haskell
(<$$>) :: (a -> b) -> K a x -> K b x
f <$$> (K a) = K (f a)

(<**>) :: K (a -> b) x -> K a y -> K b z
(K f) <**> (K a) = K $ f a
```

As we saw when discussing `ExpF`, because of type safety, all of its recursive components need to have very specific type parameters. These are determined by the shape of the AST, *not* by the recursion scheme or what we want to compute. For example, even if we want to compute an `Int`, we can't have `And :: r Int -> r Int -> ExpF r Bool`. So we can't store the information we want to compute in the type parameter, but instead we have to store it in the parameter `r :: * -> *`.

This is why we use the constant functor `K`. We saw an example where `r ~ K Text`. Then we get `And :: K Text Bool -> K Text Bool -> ExpF (K Text) Bool`. Nice, everything typechecks. And since `K String Bool` is isomorphic to `String`, we have indeed succeeded in replacing all recursive components with the result of a computation.

### 3. `h`, as in higher-order functors and foldables

Our type `ExpF` can be mapped and folded over, but not in the normal way. Because of type safety it wouldn't be possible to define a normal functor instance; it is completely nonsensical to try to apply `fmap (>5)` on a term of type `Exp Int`, even though the types *seem* to match.

What we can do instead is to define an instance of [the typeclass `HFunctor`](https://hackage.haskell.org/package/compdata-0.12.1/docs/Data-Comp-Multi-HFunctor.html#t:HFunctor). This allows us to transform the recursive parameter `r :: * -> *` to some `s :: * -> *`. For example, we could do `hfmap (K . length . unK) :: ExpF (K String) Bool -> ExpF (K Int) Bool`.

Similarly, `ExpF` has an instance of the [`HFoldable` typeclass](https://hackage.haskell.org/package/compdata-0.12.1/docs/Data-Comp-Multi-HFoldable.html#t:HFoldable), which allows us to fold over the recursive parameter. `hfold` is a special method which assumes that the parameter we are folding over is a `Monoid` embedded in `K`, as can be seen from its type:

```haskell
hfold :: (HFoldable h, Monoid m) => h (K m) i -> m
```

If this is not the case, there also exists the more general function:

```haskell
hfoldMap :: (HFoldable h, Monoid m) => (forall i. a i -> m) -> (forall i. h a i -> m) 
```

Putting this together, if we have an `ExpF (K [StorageLocation]) a` as in `locsFromExp` above, we can concatenate all the `[StorageLocation]`s which it contains into one single `[StorageLocation]` simply by using `hfold` (or `hfoldMap unK`, which is exactly how `hfold` is defined). In this case, the specialized type of `hfold` is:

```haskell
hfold :: ExpF (K [StorageLocation]) i -> [StorageLocation]
```

### 4. Recursion schemes

Finally, we've arrived. Let's have a look at `cata`, `zygo` and their siblings, starting with `cata`, or more frighteningly, *catamorphisms*.

#### Catamorphisms

Catamorphisms are more or less just fancy folds. Some other recursion schemes are even fancier folds, and yet others are fancy *un*folds and *re*folds. But one thing at a time. The type of `cata` looks like this:

```haskell
cata :: HFunctor f => Alg f a -> Term f :-> a
```

Let's specialize this to our case from `locsFromExp`:

```haskell
cata :: Alg ExpF (K [StorageLocation]) -> Term ExpF :-> K [StorageLocation]
```

This may look weird, but it's relatively straight forward:
- `ExpF`, `Term` and `K` should be familiar by now.
- `:->` is a function between functors. For example the type `forall i. Maybe i -> [i]` could also be written `Maybe :-> []`. (This is called a *natural transformation*, but this isn't necessary to remember.)
- `Alg f a` is an *algebra* of `f` with `a` as its carrier. Think of this as a folding function from `f` to `a`, which can inspect the structure of `f` instead of only "folding it like a list". Technically speaking, `Alg f a` is just a type synonym for `f a :-> a`, which, recalling the previous bullet point, is the same as `forall i. f a i -> a i`. So in this specialization, it's a way of taking an `ExpF` – in which `K [StorageLocation]` has replaced the recursive components! – and turning it into a `K [StorageLocation]`, regardless of its type parameter.
- Putting this together, `cata` takes
    - an algebra/folding function from the structure `ExpF` into the structure `K [StorageLocation]` (which assumes that it has already been applied recursively) and
    - a fully recursive term of `ExpF`.
    - It then folds this term, using the algebra, and takes care of the recursion for us.

#### Zygomorphisms

We already saw one other recursion scheme, `zygo`, which allowed us to *select* which algebra we wanted to use, depending on the pattern matching case. Let's look at its type:

```haskell
zygo :: HFunctor f => Alg f b -> GRAlg f b a -> Term f :-> a
```

Except for `GRAlg`, all of this should make sense. And as the name may suggest, `GRAlg` is just an algebra but with a slightly different shape. Specifically, it looks like this:

```haskell
type GRAlg f b a = f (b :*: a) :-> a
```

Again, this folds our flattened type `f` into some functor `a` which it contains. But it doesn't *simply* contain `a`, it contains `b :*: a`. This is a *functor product* of the functors `b` and `a`. In other words, it's a tuple, but with functors. If you want to have a look at the [definiton](https://hackage.haskell.org/package/base-4.15.0.0/docs/GHC-Generics.html#t::-42-:) and work out the types by yourself, that can definitely be illuminating. But conceptually, since we're mostly working with `K` and `K a ≅ a`, this is *just a tuple*.

So looking at `GRAlg` again, what this says in our case is: Given a flattened structure in which the recursive components have been replaced with tuples of `b` and `a`, return an `a`. And looking at `zygo`, the `b` is actually the carrier of the *first* algebra. So in the second algebra, whenever we request a `b`, the first algebra will be used to compute it. And whenever we request an `a`, we will simply compute it ourselves. The rest of `zygo` is exactly the same as `cata`.

#### Other recursion schemes

There are many different recursion schemes. I won't discuss them here since this PR is already a mile long and I still have some things left to discuss. But [this PDF gives a good overview](https://ipfs.io/ipfs/QmTppu1VDAQWsdiyVSZX6qb8PErdpwzNP2oKfEhcgaBvWR/guide-to-morphisms.pdf), if not very pedagogical. [This talk is a good intro](https://www.youtube.com/watch?v=F5NjINDNwYc) and much more pedagogical. [This flowchart can help you choose the right one for your task](https://github.com/recursion-schemes/recursion-schemes/blob/master/docs/flowchart.svg). Many additional resources can be found in the [Awesome recursion schemes repository](https://github.com/passy/awesome-recursion-schemes).

### On `:*:` and `zygoK`

That `:*:` together with `K` is "just a tuple" is all fine, but how do we destruct it to get the individual components? Simple, use `ffst` or `fsnd`. Alternatively, the data constructor is the same as the type constructor, so just pattern match on `(x:*:y)`. Both `x` and `y` will then have some types on the form `K a b`.

So why didn't we have to do this when we used `zygo`? Well, we could. But remember that we actually used `zygoK`, which takes advantage of the fact that we are returning `K` rather than just any functor, and so it uses this to get us to a `K (x,y)` instead of a `K x :*: K y`.

But wait, we didn't pattern match on `K (x,y)` either? Well, we actually did. You just didn't see it. Remember that we matched on `Fst` and `Snd`? Those aren't actually constructors, but *pattern synonyms* defined using the GHC `PatternSynonyms` extension. They're defined like this in `Utils.hs`:

```haskell
pattern Fst  a   <- K (a,_)
pattern Snd  b   <- K (_,b)
pattern Both a b <- K (a,b)
```

So matching on `Fst a` is literally the same thing as matching on `K (a,_)`. Magic!

A different way of implementing `coqprop` would be to use `zygo` instead of `zygoK`:

```haskell
coqprop' :: Exp Bool -> T.Text
coqprop' = unK . zygo (K . expAlg) \case
  LitBool b              -> K . T.pack . show $ b
  And  (_:*:e1) (_:*:e2) -> infix2  "/\\" <$$> e1 <**> e2
  Eq   (e1:*:_) (e2:*:_) -> infix2  "="   <$$> e1 <**> e2
```

In some ways, this may seem more complicated. We have to deal with some `K` and `unK` bookkeeping, and use scary-looking operators. But in some ways, it's also significantly simpler. There is less magic going on. And it doesn't require us to define (and maintain!) a bunch of utility stuff. Additionally, I suspect that if we decide that we do want to start using recursion schemes, other schemes will require similar utility magic if we want to be consistent.

# Questions

1. Does this seem at all useful/desirable?
2. Which style seems most useful? Do we want to be explicit with the `K` stuff or do we want to attempt to hide it with utility magic as far as possible? Or only hide it in some cases?


kjekac:
After first finishing the poast and then sleeping on it, my relatively strong opinion is that we do want to use recursion schemes, but that except from some `unK` and `K` noise reduction, we don't want any helper functions or syntactic sugar. The operators may look scary at first glance but they're conceptually very simple and map neatly onto things that everyone already knows.

kjekac:
I and @xwvvvvwx were talking a bit about this today. A nice middle-ground seems to be to keep the changes to the AST but only make use of it in places where it provides a clear syntactic advantage (such as `locsFromExp` or `ethEnvFromExp`). We can still recurse manually if we want, we don't *have* to replace everything.

This way we make the code nicer, we reduce the barrier to entering the codebase (fewer scary operators) *and* keep the door open to using this stuff if we want to do something more complicated in the future. David had an example where it may turn out to be beneficial but I already forgot what it was.

The disadvantages of this compromise would be
- inconsistency (but honestly who cares).
- making the AST type more complicated and introducing an external dependency, while currently only using this in two places.

src/RefinedAst.hs
```diff
@@ -37,7 +42,9 @@ data Invariant = Invariant
   { _icontract :: Id
   , _ipreconditions :: [Exp Bool]
   , _predicate :: Exp Bool
-  } deriving (Show, Eq)
+  }
+deriving instance Show Invariant
```
MrChico, line 46: why do you prefer this format?

kjekac: I don't, but with the old style we get:

```
[3 of 3] Compiling RefinedAst       ( RefinedAst.hs, interpreted )

RefinedAst.hs:45:15: error:
    • No instance for (ShowHF ExpF)
        arising from the second field of ‘Invariant’ (type ‘[Exp Bool]’)
      Possible fix:
        use a standalone 'deriving instance' declaration,
          so you can specify the instance context yourself
    • When deriving the instance for (Show Invariant)
   |
45 |   } deriving (Show, Eq)
   |               ^^^^

RefinedAst.hs:45:21: error:
    • No instance for (EqHF ExpF)
        arising from the second field of ‘Invariant’ (type ‘[Exp Bool]’)
      Possible fix:
        use a standalone 'deriving instance' declaration,
          so you can specify the instance context yourself
    • When deriving the instance for (Eq Invariant)
   |
45 |   } deriving (Show, Eq)
   |                     ^^
```

src/act.cabal
```diff
@@ -23,8 +23,11 @@ common deps
                  bytestring       >= 0.10.8,
                  sbv              >= 8.4,
                  mtl              >= 2.2.2,
-                 utf8-string      >= 1.0.1.1
-  other-modules: Lex ErrM Extract Parse RefinedAst K HEVM Coq Syntax Type Prove Print Enrich
+                 utf8-string      >= 1.0.1.1,
+                 template-haskell >= 2.14,
```
MrChico, line 27: where is template haskell used?

src/RefinedAst.hs
MrChico: Makes sense

src/act.cabal
kjekac: [When deriving instances](https://github.com/ethereum/act/blob/f7715c6cd40db9317a7ddbb71b4baf02ca226bc5/src/RefinedAst.hs#L167) of `HFunctor`, `HFoldable`, `ShowHF`, `EqHF` and creating some "constructors" (really functions) which allow us to create `Exp a`s as before.

xwvvvvwx:
Sorry for the even slower review :grin: 

Regarding this PR and the usage of recursion schemes in this codebase, it seems to be a question of taste and style (do we prefer a more complex / instructable AST type, or some extra boilerplate when working with this type). Personally, I would tend to agree with @MrChico that a little boilerplate is nothing to be scared of, and I also still find the details of the way that `Term` and `K` are used in the expression type to be rather mysterious. 

However, given that the impact of the modifications to the AST seems to be relatively light for code that is not making use of a recursion scheme (mostly we will just have to use `case unterm e of` instead of `case e or` I think?), I think I would be OK to merge a version of this PR as long as we restrict usage of the recursion schemes to places where it seems to be an obvious win (right now just `locsFromExp` / `ethEnvFromExp` afaict?).

kjekac:
>However, given that the impact of the modifications to the AST seems to be relatively light for code that is not making use of a recursion scheme (mostly we will just have to use `case unterm e of` instead of `case e or` I think?)

Yes, that or pattern match on `Term e`. You can see examples of each of those in [`toJSON`](
https://github.com/ethereum/act/blob/recursion-schemes-2/src/RefinedAst.hs#L245) and [`prettyExp`](https://github.com/ethereum/act/blob/recursion-schemes-2/src/Print.hs#L41). I'm using `fixExp` instead of `unTerm` but they are literally identical, the former is a leftover from a previous architecture which I just defined as equal to the latter. If we decide that we do want to merge I'll remove that before (unless you find the name helpful for intuition?).

However what you should keep in mind is that while the impact on *correct* code is relatively low, the impact on *incorrect* code is higher. Here are the (first few) errors I get if I forget to do `unTerm` before pattern matching:

```
Print.hs:45:3: error:
    • Couldn't match type ‘ExpF r16 t16’
                     with ‘Data.Comp.Multi.Term.Cxt
                             Data.Comp.Multi.Term.NoHole ExpF (Data.Comp.Multi.HFunctor.K ()) a’
      Expected type: Exp a
        Actual type: ExpF r16 t16
    • In the pattern: Or a b
      In a case alternative: Or a b -> print2 "or" a b
      In the expression:
        case e of
          Or a b -> print2 "or" a b
          [...]
    • Relevant bindings include
        e :: Exp a (bound at Print.hs:42:11)
        prettyExp :: Exp a -> String (bound at Print.hs:42:1)
   |
45 |   Or a b -> print2 "or" a b
   |   ^^^^^^

Print.hs:45:25: error:
    • Couldn't match type ‘r16’
                     with ‘Data.Comp.Multi.Term.Cxt
                             Data.Comp.Multi.Term.NoHole ExpF (Data.Comp.Multi.HFunctor.K ())’
        ‘r16’ is untouchable
          inside the constraints: t16 ~ Bool
          bound by a pattern with constructor:
                     Or :: forall (r :: * -> *). r Bool -> r Bool -> ExpF r Bool,
                   in a case alternative
          at Print.hs:45:3-8
      Expected type: Exp Bool
        Actual type: r16 Bool
    • In the second argument of ‘print2’, namely ‘a’
      In the expression: print2 "or" a b
      In a case alternative: Or a b -> print2 "or" a b
    • Relevant bindings include
        b :: r16 Bool (bound at Print.hs:45:8)
        a :: r16 Bool (bound at Print.hs:45:6)
   |
45 |   Or a b -> print2 "or" a b
   |                         ^
```

And here's the error I get if I try to pass an `ExpF` instead of an `Exp` to `show`.

```
RefinedAst.hs:280:26: error:
    • No instance for (Show
                         (ExpF (Cxt NoHole ExpF (Data.Comp.Multi.HFunctor.K ())) t))
        arising from a use of ‘show’
    • In the second argument of ‘(.)’, namely ‘show’
      In the second argument of ‘(.)’, namely ‘pack . show’
      In the expression: String . pack . show
    |
280 |     _ -> String . pack . show $ e
    |                          ^^^^
```

Tbh the more distance I get to this PR the more I feel that we do not want it and that we should just chalk it up as my weird way of getting familiar with the code base. I mean let's keep the branch around if we run into a more promising use case later but at the moment I think it's just not worth it. Or am I being overly critical here?

Issue #71, opened 2020-12-15
State var can be used to initialize itself


leonardoalt:
```
constructor of LValue
interface constructor()

creates

    uint x := x
    
invariants

    x == 0
```
This is accepted by parse/type/prove.

The invariant is false (any expression you put there), since I guess `x` is unconstrained initially.

I guess this should be disallowed in general?

xwvvvvwx:
I think lets just disallow storage references on the RHS of assignment expressions in the creates block for now, we can reindtroduce them later (probably with some kind of cycle detection?)

MrChico:
Yeah, that sounds reasonable. A variable on RHS in the constructor would always be empty, anyway

PR #104, opened 2021-05-05
SMT: storage references refer to the prestate

xwvvvvwx:
The intended current semantics of storage references is that they always refer to the prestate, this PR fixes the encoding of expressions on the RHS of an assignment in a `creates` block to adhere to this semantics. Note that since the prestate does not exist for the constructor, this breaks all storage refs in the `creates` block, but since we intend to make this a type error anyway (https://github.com/ethereum/act/issues/71), that seems fine.


kjekac (approved):
The changes look good, but question: What's the story with the other backends? Do they also adhere to this? Not that we necessarily have to update them in lockstep, just wondering.

xwvvvvwx:
> What's the story with the other backends? Do they also adhere to this?

Very good question, I wouldn't be surprised if similar issues exist in the other backends. I'm not sure if it's worth worrying about too much considering we're intending to introduce and enforce `pre`/`post` though?

kjekac:
Yeah, reasonable.

PR #105, opened 2021-05-07
Prevent storage read in `creates` block

### LINKED ISSUE #71: State var can be used to initialize itself

kjekac:
Closes #71 

One issue I can currently see is with the following test:

```
constructor of LValue
interface constructor()

creates

    uint x := y
    uint y := x
```

This does fail to typecheck, but unfortunately only with one error instead of two:

```
6 |     uint x := y
                  ^
Cannot read storage in creates block
```

This is because of the way the `Err` monad works. Maybe it is more designed for parsing than for type checking? Regardless this doesn't seem like it's covered by the issue I'm trying to solve here, but I guess it's worthwhile to consider while you're looking at error handling @xwvvvvwx?

### DIFF
--- a/src/Syntax.hs
+++ b/src/Syntax.hs
@@ -4,7 +4,7 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE OverloadedStrings #-}
 module Syntax where
-import Data.List (intercalate)
+import Data.List (intercalate,nub)
 import EVM.ABI (AbiType)
 import EVM.Solidity (SlotType)
 import Lex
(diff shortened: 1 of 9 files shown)


xwvvvvwx (commented):
very neat! :clap: 

If you just fix the "missing newline at end of file" stuff that github is compaining about, then this lgtm :sparkles:

kjekac:
lol, I never thought of that as a complaint, more a poorly phrased approval. x) But I guess your interpretation actually makes more sense so sure I'll adapt to their erroneous worldview.

xwvvvvwx (approved):

Issue #14, opened 2020-03-11
SMT checks of spec validity


MrChico:
Even after a simple type checking, there are a number of ways in which specs can be malformed, possibly leading to crashes in different prover back ends.

There are a few checks that I think can be implemented quite straightforwardly with the assistance of an SMT checker, for example using the [sbv](https://github.com/LeventErkok/sbv) library:

- Bound checking of expressions
Can the expression assigned to a particular type fall out of the bounds of that type? Throw warning to suggest that the expression be further constrained to be bound.

- Case exhaustiveness / redundancy
Throw warning if cases are not exhaustive, or if a case is redundant

- Array / mapping collisions
In the classic token example:
```act
     balanceOf[CALLER] => balanceOf[CALLER] - value
     balanceOf[to]     => balanceOf[to] + value
```
if `CALLER == to`, the same expression has been assigned to distinct values, and the spec doesn't make any sense. Specs like this should not be accepted.

MrChico:
Some inspiration: https://github.com/LeventErkok/sbv/blob/master/Documentation/SBV/Examples/Transformers/SymbolicEval.hs

franck44:
@MrChico I am not sure that using a direct encoding into SMTLIB syntax (this the example state_machine.{act,smt2}) is the best way to go.
SMT-solvers are used as back end solvers in many verification tools.
Let’s not re-invent the wheel: solidity may be translated into some language that has support for verification, and the translation may be proved “correct” or preserving some properties) which is easier than a direct encoding into SMTLIB syntax.

There are lots of techniques that are mature in the software verification community and it could be good to leverage them to build real trustworthy distributed systems.

Regarding the specs that are malformed, this is a valid point. But bear in mind that 
1. Verification is about making that the implementation is doing the right thing (comparing the implementation with the spec). This can be automated or mechanised (model-checking, theorem proving)
2. Validation is about checking that the spec captures the “customers’ requirements. This cannot usually be automated. The purpose of validation is to go from an possibly ambiguous and informal spec (say in plain English) to a formal (in a logical language or temporal logic of some sort).

leonardoalt:
@franck44 I'm not sure I get the first part of your comment. We're not reinventing the wheel: There are several tools out there that translate Solidity to some language which already has verification tooling. That's not the goal here. Act has nothing to do with Solidity (besides copying its storage layout).
[...]

PR #106, opened 2021-05-11
Simplify exponentiation in SMT backend

### LINKED ISSUE #99: SMT: further simplify exponentiation expressions
The SMT backend currently performs some very simple preprocessing of expressions involving exponentation (see [here](https://github.com/ethereum/act/blob/3944f4f06b73ee75b7226128b2bd01676f1ea21b/src/SMT.hs#L410)). These could be extended to allow the SMT encoding of any exponentiation where the RHS is concrete, by either:

- fully evaluating the result of the exponentation if both the RHS & LHS are concrete
- encoding the exponentiation as a chain of multiplications if the RHS is concrete, but the LHS is symbolic

We will need to write a simple evaluator for act expressions to do this properly.

kjekac:
Closes #99, i.e. we can now pass any exponentiation where the exponent is concrete to the SMT backend even though it has no support for exponentiation. This does not however include any algebraic analysis to attempt to simplify away symbolic values.

I'm a bit uncertain on how division is to be dealt with. Currently I'm using `div` and `mod` from Haskell's Prelude (rather than `quot` and `rem`). This is because they truncate towards negative infinity, which is what the EVM seems to be doing according to the yellow paper. But what exactly are we trying to agree with? The EVM? In that case `x/0` should be `0`, so we need a special case for that. Or Solidity? Or our other backends?

### DIFF
--- a/src/RefinedAst.hs
+++ b/src/RefinedAst.hs
@@ -22,6 +22,7 @@ import Data.Typeable
 import Data.Map.Strict (Map)
 import Data.List.NonEmpty hiding (fromList)
 import Data.ByteString (ByteString)
+import qualified Data.ByteString as BS
 import Data.String (fromString)
 
 import EVM.Solidity (SlotType(..))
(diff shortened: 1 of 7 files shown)


src/RefinedAst.hs
```diff
@@ -215,6 +220,51 @@ data ReturnExp
   | ExpBytes  (Exp ByteString)
   deriving (Eq, Show)
 
+-- | Simplifies concrete expressions into literals.
+-- Returns `Nothing` if the expression contains symbols.
+eval :: Exp a -> Maybe a
+eval e = case e of
+  And  a b    -> [a' && b' | a' <- eval a, b' <- eval b]
```
MrChico, line 227: monad comprehension! Cool, haven't seen this before

kjekac:
> > But what exactly are we trying to agree with? The EVM? In that case x/0 should be 0, so we need a special case for that. Or Solidity? Or our other backends?
> 
> IMO, specs should conform to "mathematical rigor", and not so much evm details, meaning I think we should ideally disallow specs which admit division by zero as part of future spec validity checking (discussed at #14).

Ok so following that reasoning we want to truncate downwards instead of towards zero, since that's how Euclidean division works. Then the code here should be correct.

src/SMT.hs
```diff
@@ -406,10 +409,12 @@ expToSMT2 ctx@(Ctx behvName whn) e = case e of
 
 -- | SMT2 has no support for exponentiation, but we can do some preprocessing
 --   if the RHS is concrete to provide some limited support for exponentiation
---   TODO: support any exponentiation expression where the RHS evaluates to a concrete value
 simplifyExponentiation :: Exp Integer -> Exp Integer -> Exp Integer
-simplifyExponentiation (LitInt a) (LitInt b) = LitInt (a ^ b)
-simplifyExponentiation _ _ = error "Internal Error: exponentiation is unsupported in SMT lib"
+simplifyExponentiation a b = fromMaybe (error "Internal Error: no support for symbolic exponents in SMT lib")
+                             $   [LitInt $ a' ^ b'                      | a' <- eval a,   b' <- evalb]
+                             <|> [foldr Mul (LitInt 1) (replicate b' a) | b' <- fromInteger <$> evalb]
```
xwvvvvwx, line 415: What will happen with the `fromInteger` here if the exponent is larger than the bounds of the `Int` type?

xwvvvvwx: seems like it will just return 0?

```
Prelude> let a = (2 ^ 256) :: Integer
Prelude> (fromInteger a) :: Int
0
```

xwvvvvwx: definately an edge case, but I would prefer to either keep everything as an `Integer` here or hard fail if we overflow the `Int` type.

xwvvvvwx (approved):

PR #107, opened 2021-05-11
SMT: Model Extraction & Pretty Counterexamples

### LINKED ISSUE #63: Pretty Print Counterexamples
The counterexamples are currently not pretty printed in any way, and look very messy. For exmaple for [this spec](https://github.com/ethereum/act/blob/b78ffd6715323dd93ee36aaf9d6035963e3e80f0/tests/invariants/fail/non-inductive.act):

```

Invariant "(C.x < 9)" of "C":



---



---



---

Counter example found!

SMTModel {modelObjectives = [], modelBindings = Nothing, modelAssocs = [("C_j_x",7 :: Integer),("C_j_x_post",100 :: Integer),("C_j_CALLER",0 :: Integer),("C_j_CALLVALUE",0 :: Integer),("C_j_CALLDEPTH",0 :: Integer),("C_j_ORIGIN",0 :: Integer),("C_j_BLOCKHASH","" :: String),("C_j_BLOCKNUMBER",0 :: Integer),("C_j_DIFFICULTY",0 :: Integer),("C_j_CHAINID",0 :: Integer),("C_j_GASLIMIT",0 :: Integer),("C_j_COINBASE",0 :: Integer),("C_j_TIMESTAMP",0 :: Integer),("C_j_THIS",0 :: Integer),("C_j_NONCE",0 :: Integer)], modelUIFuns = []}
```

We should tidy this up, perhaps something like the following:

```
Invariant C.x < 9 of C:

  Counterexample:

    prestate:

      calldata:

        j()
      
      storage:

        x = 7

      environment:

        THIS      = 0
        NONCE     = 0
        ORIGIN    = 0
        CHAINID   = 0
        GASLIMIT  = 0
        COINBASE  = 0
        TIMESTAMP = 0
        CALLVALUE = 0
        CALLDEPTH = 0

    poststate:

      x = 100
```

xwvvvvwx:
This closes #63.

Notable changes:

- Pretty printed counterexamples
- Reworked solver interaction code so that we now share a single long running solver instance between all queries. This is a little more involved, but makes extracting the models much easier
- The `Inv` constructor for the `Query` type has been reworked a little to bundle all the individual smt queries related to each invariant in one type
- The various `locsFrom*` / `ethEnvFrom*` functions in `Extract.hs` now deduplicate their output, making them more ergononomic at the cost of some redundent calls to `nub`

Pretty printing / colorful output is achieved using the [`ansi-wl-pprint`](https://hackage.haskell.org/package/ansi-wl-pprint-0.6.7.3/docs/Text-PrettyPrint-ANSI-Leijen.html#v:encloseSep) library. It may be worth reworking the pretty printer in `Print.hs` as well as the coq and k backends to make use of this, but I've left that for another day.

As I worked through this I realised that error handling in the SMT backend is also a little unsatisfactory, but since this is already quite a lot of code to review I've decided to tidy that up properly in another PR.

An example of a pretty printed counterexample:

```
    counterexample:

      calldata:

        f(x : 1, y : 0, z : true)

      storage:

        prestate:

          C.a : 0
          C.c : 0
          C.b : true

        poststate:

          C.a : 1
          C.c : 1
          C.b : false

      constructor arguments:

        constructor(x : 0, y : false, z : "")
```

### DIFF
--- a/src/Extract.hs
+++ b/src/Extract.hs
@@ -3,27 +3,28 @@
 module Extract where
 
 import qualified Data.List.NonEmpty as NonEmpty
+import Data.List
 
 import RefinedAst
 import Syntax
 import EVM.ABI (AbiType(..))
 
 locsFromBehaviour :: Behaviour -> [StorageLocation]
-locsFromBehaviour (Behaviour _ _ _ _ preconds postconds stateUpdates returns) =
+locsFromBehaviour (Behaviour _ _ _ _ preconds postconds stateUpdates returns) = nub $
   (concatMap locsFromExp preconds)
   <> (concatMap locsFromExp postconds)
   <> (concatMap locsFromStateUpdate stateUpdates)
   <> (maybe [] locsFromReturnExp returns)
 
 locsFromConstructor :: Constructor -> [StorageLocation]
-locsFromConstructor (Constructor _ _ _ pre post initialStorage stateUpdates) =
+locsFromConstructor (Constructor _ _ _ pre post initialStorage stateUpdates) = nub $
   (concatMap locsFromExp pre)
   <> (concatMap locsFromExp post)
   <> (concatMap locsFromStateUpdate stateUpdates)
   <> (concatMap locsFromStateUpdate (Right <$> initialStorage))
 
 locsFromStateUpdate :: Either StorageLocation StorageUpdate -> [StorageLocation]
-locsFromStateUpdate update = case update of
+locsFromStateUpdate update = nub $ case update of
   Left loc -> [loc]
   Right (IntUpdate item e) -> (IntLoc item) : locsFromExp e
   Right (BoolUpdate item e) -> (BoolLoc item) : locsFromExp e
@@ -35,7 +36,7 @@ locsFromReturnExp (ExpBool e) = locsFromExp e
 locsFromReturnExp (ExpBytes e) = locsFromExp e
 
 locsFromExp :: Exp a -> [StorageLocation]
-locsFromExp e = case e of
+locsFromExp e = nub $ case e of
   And a b   -> (locsFromExp a) <> (locsFromExp b)
   Or a b    -> (locsFromExp a) <> (locsFromExp b)
   Impl a b  -> (locsFromExp a) <> (locsFromExp b)
@@ -81,21 +82,21 @@ locsFromExp e = case e of
       ixLocs = concatMap locsFromReturnExp
 
 ethEnvFromBehaviour :: Behaviour -> [EthEnv]
-ethEnvFromBehaviour (Behaviour _ _ _ _ preconds postconds stateUpdates returns) =
+ethEnvFromBehaviour (Behaviour _ _ _ _ preconds postconds stateUpdates returns) = nub $
   (concatMap ethEnvFromExp preconds)
   <> (concatMap ethEnvFromExp postconds)
   <> (concatMap ethEnvFromStateUpdate stateUpdates)
   <> (maybe [] ethEnvFromReturnExp returns)
 
 ethEnvFromConstructor :: Constructor -> [EthEnv]
-ethEnvFromConstructor (Constructor _ _ _ pre post initialStorage stateUpdates) =
+ethEnvFromConstructor (Constructor _ _ _ pre post initialStorage stateUpdates) = nub $
   (concatMap ethEnvFromExp pre)
   <> (concatMap ethEnvFromExp post)
   <> (concatMap ethEnvFromStateUpdate stateUpdates)
   <> (concatMap ethEnvFromStateUpdate (Right <$> initialStorage))
 
 ethEnvFromStateUpdate :: Either StorageLocation StorageUpdate -> [EthEnv]
-ethEnvFromStateUpdate update = case update of
+ethEnvFromStateUpdate update = nub $ case update of
   Left (IntLoc item) -> ethEnvFromItem item
   Left (BoolLoc item) -> ethEnvFromItem item
   Left (BytesLoc item) -> ethEnvFromItem item
@@ -104,7 +105,7 @@ ethEnvFromStateUpdate update = case update of
   Right (BytesUpdate item e) -> ethEnvFromItem item <> ethEnvFromExp e
 
 ethEnvFromItem :: TStorageItem a -> [EthEnv]
-ethEnvFromItem item = case item of
+ethEnvFromItem item = nub $ case item of
   MappedInt _ _ ixs -> concatMap ethEnvFromReturnExp ixs
   MappedBool _ _ ixs -> concatMap ethEnvFromReturnExp ixs
   MappedBytes _ _ ixs -> concatMap ethEnvFromReturnExp ixs
@@ -116,7 +117,7 @@ ethEnvFromReturnExp (ExpBool e) = ethEnvFromExp e
 ethEnvFromReturnExp (ExpBytes e) = ethEnvFromExp e
 
 ethEnvFromExp :: Exp a -> [EthEnv]
-ethEnvFromExp e = case e of
+ethEnvFromExp e = nub $ case e of
   And a b   -> ethEnvFromExp a <> ethEnvFromExp b
   Or a b    -> ethEnvFromExp a <> ethEnvFromExp b
   Impl a b  -> ethEnvFromExp a <> ethEnvFromExp b
--- a/src/Extract.hs
+++ b/src/Extract.hs
@@ -86,16 +86,16 @@ locsFromExp = nub . go
 
 ethEnvFromBehaviour :: Behaviour -> [EthEnv]
 ethEnvFromBehaviour (Behaviour _ _ _ _ preconds postconds stateUpdates returns) = nub $
-  (concatMap ethEnvFromExp preconds)
-  <> (concatMap ethEnvFromExp postconds)
-  <> (concatMap ethEnvFromStateUpdate stateUpdates)
-  <> (maybe [] ethEnvFromReturnExp returns)
+  concatMap ethEnvFromExp preconds
+  <> concatMap ethEnvFromExp postconds
+  <> concatMap ethEnvFromStateUpdate stateUpdates
+  <> maybe [] ethEnvFromReturnExp returns
 
 ethEnvFromConstructor :: Constructor -> [EthEnv]
 ethEnvFromConstructor (Constructor _ _ _ pre post initialStorage stateUpdates) = nub $
-  (concatMap ethEnvFromExp pre)
-  <> (concatMap ethEnvFromExp post)
-  <> (concatMap ethEnvFromStateUpdate stateUpdates)
+  concatMap ethEnvFromExp pre
+  <> concatMap ethEnvFromExp post
+  <> concatMap ethEnvFromStateUpdate stateUpdates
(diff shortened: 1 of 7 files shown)


src/test/Test.hs
```diff
@@ -78,12 +78,18 @@ typeCheckSMT :: Solver -> GenT (Reader Bool) Property
...
-      parseOutput (exitCode, stdout, _) = case exitCode of
-        ExitFailure _ -> False
-        ExitSuccess -> not . any (isPrefixOf "(error") . filter (/= "") . lines $ stdout
+      runQueries smtconf queries = do
+        solverInstance <- spawnSolver smtconf
+        res <- mapM (askSMT solverInstance) queries
+        pure $ null (catMaybes res)
+
+      getSMT (Postcondition _ _ smt) = smt
+      getSMT _ = undefined
```
kjekac, line 90: Even though the input will always have been run through `mkPostconditionQueries` I would say that using `undefined` is bad practice in production code. (And `mkPostconditionQueries` may not do what the name suggests.) Maybe at least use `error "some nice msg"` instead?

src/SMT.hs
```diff
@@ -96,7 +156,7 @@ data Model = Model
 --    - Asserts that the postcondition cannot be reached
 --   If this query is unsatisfiable, then there exists no case where the postcondition can be violated.
 mkPostconditionQueries :: Claim -> [Query]
-mkPostconditionQueries (B behv@(Behaviour _ _ _ (Interface ifaceName decls) preconds postconds stateUpdates _)) = mkQuery <$> postconds
+mkPostconditionQueries (B behv@(Behaviour _ Pass _ (Interface ifaceName decls) preconds postconds stateUpdates _)) = mkQuery <$> postconds
```
kjekac, line 159: Why do we care about `Pass`?

src/Main.hs
```diff
@@ -105,25 +107,63 @@ main = do
         contents <- readFile file'
         proceed contents (compile contents) $ \claims -> do
           let
-            handleRes (query, res) = case res of
-                  Unsat -> (True, typ <> " " <> prettyExp target <> " holds :)", show $ getSMT query)
-                  Sat -> (False, typ <> " " <> prettyExp target <> " does not hold :(", show $ getSMT query)
-                  SMT.Unknown -> (False, typ <> " " <> prettyExp target <> " could not be proved due to a solver timeout :(", show $ getSMT query)
-                  SMT.Error _ str -> (False, typ <> " " <> prettyExp target <> " could not be proved to due a solver error: " <> str, show $ getSMT query)
+            catModels results = [m | Sat m <- results]
```
kjekac, line 110: Maybe this is just me but I find this amount of nested `do`s, `let`s and `where`s to be quite difficult to follow. Many of these functions seem to be directly related to printing SMT stuff, and so maybe could be defined in `SMT.hs` instead?

src/Extract.hs
```diff
@@ -3,18 +3,40 @@
...
+  Right (BytesUpdate item e) -> (BytesLoc item) : locsFromExp e
+
 locsFromReturnExp :: ReturnExp -> [StorageLocation]
 locsFromReturnExp (ExpInt e) = locsFromExp e
 locsFromReturnExp (ExpBool e) = locsFromExp e
 locsFromReturnExp (ExpBytes e) = locsFromExp e
 
 locsFromExp :: Exp a -> [StorageLocation]
-locsFromExp e = case e of
+locsFromExp e = nub $ case e of
```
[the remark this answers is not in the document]
xwvvvvwx: nice, but `nub` is actually `O(n^2)` :(

We can get `O(n log(n))` if we have an `ord` instance for `Exp`.

src/SMT.hs
xwvvvvwx: This was a bug, postconditions should not hold for `Fail` queries (and actually thinking about it now, probably we shouldn't even include them when we split the act into pass / fail specs?).

src/Print.hs
```diff
@@ -1,6 +1,6 @@
 {-# Language GADTs #-}
 
-module Print (prettyEnv, prettyExp, prettyType, prettyBehaviour, prettyItem) where
+module Print where
```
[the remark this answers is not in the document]
xwvvvvwx: well in the end I just thought that we wanted everything in the module to be exported and I was annoyed by that I kept having to add functions to the export list here.

Is there some agreed upon best practice for module import / export lists? We have quite a mix in our codebase at the moment...

src/Main.hs
xwvvvvwx: Tidied up a bit here, think it should be a little clearer now.

xwvvvvwx:
addressed review comments and rebased on master

src/Print.hs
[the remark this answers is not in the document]
MrChico: I always found the explicit export function and import function kinda verbose and beaurocratic

src/SMT.hs
```diff
@@ -59,36 +82,75 @@ data SMTExp = SMTExp
...
+      calldata' = (text "calldata:") <$$> line <> (indent 2 $ formatSig ifaceName args)
+      environment' = (text "environment:") <$$> line <> (indent 2 . vsep $ fmap formatEnvironment environment)
+      storage = (text "storage:") <$$> (indent 2 . vsep $ [ifExists prestate (line <> prestate'), poststate'])
+      initargs' = (text "constructor arguments:") <$$> line <> (indent 2 $ formatSig "constructor" initargs)
+
+      prestate' = (text "prestate:") <$$> line <> (indent 2 . vsep $ fmap formatStorage prestate) <> line
+      poststate' = (text "poststate:") <$$> line <> (indent 2 . vsep $ fmap formatStorage poststate)
+
+      formatSig iface cd = (text iface) <> (encloseSep lparen rparen (text ", ") $ fmap formatCalldata cd)
+      formatCalldata ((Decl _ name), val) = text $ name <> " : " <> prettyReturnExp val
```
MrChico, line 142: I'd maybe go for `=` instead of `:` here. I associate `:` with typing

src/Main.hs
```diff
@@ -105,25 +106,49 @@ main = do
...
-
-          pcResults <- mapM (runQuery config) (concatMap mkPostconditionQueries claims)
-          invResults <- mapM (runQuery config) (mkInvariantQueries claims)
-          let results = map handleRes (pcResults <> invResults)
-          allGood <- foldM (\acc (r, msg, smt) -> do
-            if (_debug config) then putStrLn (msg <> "\n\n" <> smt) else putStrLn msg
-            pure $ if acc == False then False else r
-            ) True results
-          unless allGood exitFailure
+                holds = and (fmap isPass results)
```
kjekac, line 128:
```suggestion
                holds = all isPass results
```

src/Main.hs
```diff
@@ -105,25 +106,49 @@ main = do
...
+                msg = identifier query <+> if holds then passMsg else failMsg results
+                smt = if debug' then line <> (getSMT query) else empty
```
kjekac, line 130:
```suggestion
                smt = if debug' then line <> getSMT query else empty
```

kjekac (commented):
I unfortunately still haven't gone through `SMT.hs` properly but here are a few other things I found.

src/Main.hs
kjekac: Yes, much better!

src/Print.hs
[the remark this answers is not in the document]
kjekac: I've no idea lol. I think it makes sense to restrict exports when you're building EDSLs or datatypes which should have very specific interfaces but on our case I think it's fine to just export everything most of the time.

src/SMT.hs
kjekac: I'd change it to `=` for environment and storage too, for the same reason.

src/SMT.hs
```diff
@@ -59,36 +82,75 @@ data SMTExp = SMTExp
...
+    where
+      calldata' = (text "calldata:") <$$> line <> (indent 2 $ formatSig ifaceName args)
+      environment' = (text "environment:") <$$> line <> (indent 2 . vsep $ fmap formatEnvironment environment)
+      storage = (text "storage:") <$$> (indent 2 . vsep $ [ifExists prestate (line <> prestate'), poststate'])
+      initargs' = (text "constructor arguments:") <$$> line <> (indent 2 $ formatSig "constructor" initargs)
+
+      prestate' = (text "prestate:") <$$> line <> (indent 2 . vsep $ fmap formatStorage prestate) <> line
+      poststate' = (text "poststate:") <$$> line <> (indent 2 . vsep $ fmap formatStorage poststate)
+
+      formatSig iface cd = (text iface) <> (encloseSep lparen rparen (text ", ") $ fmap formatCalldata cd)
```
kjekac, line 141:
```suggestion
      calldata' = text "calldata:" <$$> line <> (indent 2 $ formatSig ifaceName args)
      environment' = text "environment:" <$$> line <> (indent 2 . vsep $ fmap formatEnvironment environment)
      storage = text "storage:" <$$> (indent 2 . vsep $ [ifExists prestate (line <> prestate'), poststate'])
      initargs' = text "constructor arguments:" <$$> line <> (indent 2 $ formatSig "constructor" initargs)

      prestate' = text "prestate:" <$$> line <> (indent 2 . vsep $ fmap formatStorage prestate) <> line
      poststate' = text "poststate:" <$$> line <> (indent 2 . vsep $ fmap formatStorage poststate)

      formatSig iface cd = text iface <> (encloseSep lparen rparen (text ", ") $ fmap formatCalldata cd)
```

kjekac (commented):
This looks great! :) Some of my suggestions are definitely nitpicks and with some exceptions, I think most aren't about correctness or even best practices, but more about style. But I figured you'd appreciate to at least have seen those even if you don't want to apply them?

Other than that, I'd suggest that we start adding comments/documentation to any functions we touch in a PR? I believe I've done so in my last two PRs and intend to continue. Would you mind adding some here too? 😬

src/Extract.hs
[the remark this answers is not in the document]
kjekac: Damn, of course. Let's deal with that later, maybe? One single `O(n^2)` seems fine for now.

src/SMT.hs
kjekac: Oh and you also don't need the parentheses here 🙃: 

```suggestion
      formatCalldata (Decl _ name, val) = text $ name <> " : " <> prettyReturnExp val
```

src/SMT.hs
```diff
@@ -224,42 +286,226 @@ mkInvariantQueries claims = concatMap mkQueries gathered
...
+  _ <- sendCommand solverInstance "(set-option :print-success true)"
+  err <- sendLines solverInstance smtPreamble
+  case err of
+    Nothing -> pure solverInstance
+    Just msg -> error $ "could not spawn solver: " <> msg
+
+stopSolver :: SolverInstance -> IO ()
+stopSolver (SolverInstance _ stdin stdout stderr process) = cleanupProcess (Just stdin, Just stdout, Just stderr, process)
+
+sendLines :: SolverInstance -> [SMT2] -> IO (Maybe String)
```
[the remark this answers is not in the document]
xwvvvvwx: yeah, this is definately weird, and should definately be fixed by the error handling rework.

kjekac (approved):
I'm into this. :) Comments would still be good (guessing you're working on those atm?) but the code itself seems good to go imo!

kjekac:
This looks great to my mind :)

PR #115, opened 2021-07-03
represent bytes32 as an int internally

d-xo:
bytestrings are not well supported in act at the moment so this lets us verify contracts involving `bytes32` in hevm and Coq, and it should anyway be a nice optimisation.


kjekac (approved):
Looks good to me :)

kjekac (changes requested):
Sorry I meant to do this

kjekac (commented):
Great :)

kjekac (approved):

PR #119, opened 2021-07-21
main: seperate top level cli functions to allow easy repl driven workflows

d-xo:
Just a lil quality of life improvement when using the repl, allows calling each of the top level cli endpoints directly in the repl...


kjekac (approved):
Looks great, I like this!

PR #123, opened 2021-08-08
Fix display of invariant predicates

d-xo:
We currently have a little leakage of our internal representation into the ui, this PR prints a timed `InvariantPred` in a way that should be more easily digestible for end users.

Before:

```
> act prove --file src/amm.act

Invariants:

  (((1000 * 1000) <= (pre(Amm.reserve0) * pre(Amm.reserve1))) and ((1000 * 1000) <= (post(Amm.reserve0) * post(Amm.reserve1)))) of Amm holds ∎
```

After:

```
> act prove --file src/amm.act

Invariants:

  ((1000 * 1000) <= (Amm.reserve0 * Amm.reserve1)) of Amm holds ∎
```


src/Syntax.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE DataKinds #-}
```
kjekac, line 2: Would be unnecessary here since nothing else changes in this file?

```suggestion
```

src/Print.hs
```diff
@@ -122,3 +126,59 @@ prettyEnv e = case e of
...
+prettyInvPred :: InvariantPred Timed -> String
+prettyInvPred = prettyExp . stripTime . fst
+  where
+    stripTimeTyped :: Annotated.TypedExp -> Agnostic.TypedExp Untimed
+    stripTimeTyped (ExpInt e) = ExpInt (stripTime e)
+    stripTimeTyped (ExpBool e) = ExpBool (stripTime e)
+    stripTimeTyped (ExpBytes e) = ExpBytes (stripTime e)
+
+    -- | Strip timing from an annotated expression, sometimes useful for display in the ui
+    stripTime :: Annotated.Exp a -> Agnostic.Exp a Untimed
```
kjekac, line 146:
```suggestion
    stripTime :: Exp a t -> Exp a Untimed
```

src/Print.hs
```diff
@@ -9,6 +10,9 @@ import Data.List
 import Syntax
 import Syntax.TimeAgnostic
 
+import qualified Syntax.Annotated as Annotated
+import qualified Syntax.TimeAgnostic as Agnostic
```
kjekac, line 14: We don't need to mention `Annotated` here, we can just say `Exp Timed` instead. This module already knows the internal representation.

```suggestion
```

src/Print.hs
```diff
@@ -122,3 +126,59 @@ prettyEnv e = case e of
...
+-- expressions, one over the prestate and one over the poststate.  This is good
+-- since it keeps untimed expressions away from the various backends, and
+-- maintains a nice seperation between the various compilation passes, but
+-- unfortunately requires us to strip the timing out if we want to print the
+-- invariant in a way that is easily digestible by humans, requiring a less
+-- elegant implementation here than might be hoped for...
+prettyInvPred :: InvariantPred Timed -> String
+prettyInvPred = prettyExp . stripTime . fst
+  where
+    stripTimeTyped :: Annotated.TypedExp -> Agnostic.TypedExp Untimed
```
kjekac, line 140:
```suggestion
    stripTimeTyped :: TypedExp t -> TypedExp Untimed
```

src/SMT.hs
```diff
@@ -1,6 +1,7 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE MonadComprehensions #-}
+{-# LANGUAGE DataKinds #-}
```
kjekac, line 4: Is this needed here? No type signatures change.

```suggestion
```

kjekac (changes requested):

src/Print.hs
```diff
@@ -122,3 +124,58 @@ prettyEnv e = case e of
...
+prettyInvPred = prettyExp . stripTime . fst
+  where
+    stripTimeTyped :: TypedExp t -> TypedExp Untimed
+    stripTimeTyped (ExpInt e) = ExpInt (stripTime e)
+    stripTimeTyped (ExpBool e) = ExpBool (stripTime e)
+    stripTimeTyped (ExpBytes e) = ExpBytes (stripTime e)
+
+    -- | Strip timing from an annotated expression, sometimes useful for display in the ui
+    stripTime :: Exp a t -> Exp a Untimed
+    stripTime e = case e of
```
kjekac, line 145: How about changing `stripTime` to `untime`? Or even `obscure` (as opposed to `annotate` lol sorry)?

kjekac (approved):
Looks great if the tests pass (and I don't see why they shouldn't)

PR #127, opened 2021-11-18
Add source file positions to typed expressions

kjekac:
This changes absolutely no behavior, but makes it possible for the various backends to generate error messages which include a pointer to the source file position corresponding to the error. It will be used when unifying error handling.


d-xo (approved):
looks good :)

PR #131, opened 2022-03-23
This checks whether we have two behaviours with the same name

msooseth:
Only an error if it's in the same contract


msooseth:
Also fixes issues with only showing the first issue

d-xo:
CI is failing because we forgot to update the expected results in the parser tests after adding the position to the parsed transitions

d-xo:
You can cherry-pick https://github.com/ethereum/act/commit/bf6c1c2c9ab8bc66afdaf93ada16cdd88f660eff and CI should pass

PR #132, opened 2022-03-23
Checking for multiple interfaces

msooseth:
Adding test + fixing up test


tests/frontend/fail/two_interfaces.act
```diff
@@ -0,0 +1,32 @@
+// This is wrong because we have two behviours with the same name
```
d-xo, line 1:
```suggestion
// This is wrong because we have two behviours for the same interface
```

src/Type.hs
```diff
@@ -52,10 +53,17 @@ typecheck behvs = (S store:) . concat <$> traverse (splitBehaviour store) behvs
     noDuplicateVars (U.Creates assigns) = noDuplicates (fmap fst . fromAssign <$> assigns)
                            $ \x -> "Multiple definitions of Variable " <> x
 
+    noDuplicateInterfaces :: Err ()
+    noDuplicateInterfaces =
+      noDuplicates
+        [(pn, (contract ++ "." ++ (show iface))) | U.Transition pn _ contract iface _ _ _ <- behvs]
```
d-xo, line 59:
```suggestion
        [(pn, contract ++ "." ++ (show iface)) | U.Transition pn _ contract iface _ _ _ <- behvs]
```

src/Type.hs
```diff
@@ -52,10 +53,17 @@ typecheck behvs = (S store:) . concat <$> traverse (splitBehaviour store) behvs
...
+        $ \c -> "Multiple definitions of Interface " <> c
+
     noDuplicateBehaviourNames :: Err ()
     noDuplicateBehaviourNames =
-        noDuplicates [(pn, (contract ++ "." ++ behav)) | U.Transition pn behav contract _ _ _ _ <- behvs]
-                           $ \c -> "Multiple definitions of Behaviour " <> c
+      noDuplicates
+        [(pn, (contract ++ "." ++ behav)) | U.Transition pn behav contract _ _ _ _ <- behvs]
```
d-xo, line 65:
```suggestion
        [(pn, contract ++ "." ++ behav) | U.Transition pn behav contract _ _ _ _ <- behvs]
```

src/Type.hs
msooseth: Fixed, thanks!

msooseth:
:+1:

msooseth:
Thanks for all the suggested updates, helps with learning the syntax! :)

PR #133, opened 2022-03-29
init act book

d-xo:
Upgrades docs, mostly with the content from the [0.1 blog post](https://fv.ethereum.org/2021/08/31/act-0.1/), but also adds an internals section.

The docs are produced using `mdbook` and should be automatically published to github pages on every push to `master`.


msooseth:
Oops sorry I didn't review, I didn't see the mail :S

PR #135, opened 2022-08-11
Environment variable support for Coq

zoep:
Features and bug fixes: 
- Adds support for environment values in the Coq backend (only CALLER atm)
- Adds support for multiple updates to state variables 
- Bug fix in map initialization in constructors (now using default value instead of undefined var) 

Tests: 
- Added a simple token test spec and proved that sum of balances remains constant

### DIFF
--- a/lib/ActLib.v
+++ b/lib/ActLib.v
@@ -17,7 +17,7 @@ Record Env : Set :=
     Gaslimit : Z;
     Coinbase : address;
     Chainid : Z;
-    This : Z;
+    This : address;
     Origin : address;
     Nonce : Z;
     Calldepth : Z }.
--- a/lib/ActLib.v
+++ b/lib/ActLib.v
@@ -6,6 +6,10 @@ Open Scope Z_scope.
 (** * type definitions *)
 Definition address := Z.
 
+(** * Environment record *)
+Record Env : Set :=
+  { CALLER : address }.
+
 (** * integer bounds *)
 Definition UINT_MIN (n : Z) := 0.
 Definition UINT_MAX (n : Z) := 2^n - 1.
--- a/lib/ActLib.v
+++ b/lib/ActLib.v
@@ -8,7 +8,19 @@ Definition address := Z.
 
 (** * Environment record *)
 Record Env : Set :=
-  { CALLER : address }.
+  { Callvalue : Z;
+    Caller : address;
+    Blockhash : Z;
+    Blocknumber : Z;
+    Difficulty : Z;
+    Timestamp : Z;
+    Gaslimit : Z;
+    Coinbase : address;
+    Chainid : Z;
+    This : Z;
+    Origin : address;
+    Nonce : Z;
+    Calldepth : Z }.
 
 (** * integer bounds *)
 Definition UINT_MIN (n : Z) := 0.
@@ -57,4 +69,3 @@ Proof.
   - left. reflexivity.
   - right. reflexivity.
 Qed.
-
--- a/src/Coq.hs
+++ b/src/Coq.hs
@@ -292,10 +292,21 @@ coqexp (ITE _ b e1 e2) = parens $ "if "
                                <> coqexp e2
 
 -- environment values
-coqexp (IntEnv _ Caller) = parens (callerVar <> " " <> envVar)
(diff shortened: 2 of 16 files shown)


lib/ActLib.v
```diff
@@ -6,6 +6,10 @@ Open Scope Z_scope.
 (** * type definitions *)
 Definition address := Z.
 
+(** * Environment record *)
+Record Env : Set :=
+  { CALLER : address }.
```
d-xo, line 11: Can we add the full environment here:

- https://github.com/ethereum/act/blob/36b65b71eea22ebe4287f066190ecf47f4394fa3/src/Syntax/Untyped.hs#L121
- https://github.com/ethereum/act/blob/36b65b71eea22ebe4287f066190ecf47f4394fa3/src/Type.hs#L108

src/Coq.hs
```diff
@@ -158,30 +162,43 @@ stateval
...
+-- | Returns the updated value of a storage variable 
+updateVar :: [StorageUpdate]
+          -> (Id -> SlotType -> T.Text)
+          -> (Id, SlotType)
+          -> T.Text
+updateVar updates handler (name, t@(StorageValue _)) = 
+  parens $ foldl updatedVal (handler name t) (filter (eqName name) updates) where
+
+  updatedVal _ (Update SByteStr _ _) = error "bytestrings not supported"
+  updatedVal _ (Update _ _ e) = coqexp e
```
d-xo, line 180: Just a v small stylistic nit, but I prefer to keep the where on the line below, smth like this:

```suggestion
  parens $ foldl updatedVal (handler name t) (filter (eqName name) updates) 
    where
      updatedVal _ (Update SByteStr _ _) = error "bytestrings not supported"
      updatedVal _ (Update _ _ e) = coqexp e
```

src/Coq.hs
```diff
@@ -158,30 +162,43 @@ stateval
...
+
+updateVar updates handler (name, t@(StorageMapping xs _)) = parens $
+  lambda n <> foldl updatedMap prestate (filter (eqName name) updates) where
```
d-xo, line 183: same thing with the `where` here

tests/coq/token/token.act
```diff
@@ -0,0 +1,39 @@
...
+
+
+case CALLER == to:
+
+  returns 1
+
+  // storage
+
+  // 	balances[CALLER]
+  // 	balances[to]
```
d-xo, line 39: Can we either remove this or uncomment it if it's needed?

d-xo (commented):
Looks great! Just a few v small nits.

tests/coq/token/token.act
zoep: Fixed!

src/Coq.hs
zoep: Fixed

src/Coq.hs
zoep: and fixed!

lib/ActLib.v
zoep: Added the full eth env!

https://github.com/ethereum/act/blob/fd56ea17c1809a6ae52b5e250a01997919f6af9c/lib/ActLib.v#L10-L23

I tried to preserve the Z/address distinction, though I'm not quite sure what's the right choice for some fields (in particular for `this`).

https://github.com/ethereum/act/blob/fd56ea17c1809a6ae52b5e250a01997919f6af9c/src/Coq.hs#L295-L307

This could be replaced with the less verbose `coqexp (IntEnv _ envVal) = parens ((T.pack . show) envVal <> " " <> envVar)`

But the latter relies on the assumption that the Coq record field names will always be the same as the Haskell constructors. WDYT?

d-xo: `this` should be an `address`

d-xo: > This could be replaced with the less verbose coqexp (IntEnv _ envVal) = parens ((T.pack . show) envVal <> " " <> envVar)
> 
> But the latter relies on the assumption that the Coq record field names will always be the same as the Haskell constructors. WDYT?

I think that's probably fine.

d-xo (approved):
lgtm!

PR #139, opened 2023-03-21
New `checkExpr` for Act

zoep:
This PR rewrites the type checker so that it doesn't rely on Haskell type introspection (i.e., eqT). Since we plan to add custom contract types to the language this approach wouldn't let us type check contract expressions. 

It also redefines some of the Act type machinery in Syntax/Types.hs to use data kinds instead of Haskell types as indices.

### DIFF
--- a/src/Syntax/TimeAgnostic.hs
+++ b/src/Syntax/TimeAgnostic.hs
@@ -141,8 +141,8 @@ data StorageLocation (t :: Timing) where
   Loc :: SType a -> TStorageItem a t -> StorageLocation t
 deriving instance Show (StorageLocation t)
 
-_Loc :: forall a t. SingI a => TStorageItem a t -> StorageLocation t
-_Loc item = Loc (sing @a) item
+_Loc :: TStorageItem a t -> StorageLocation t
+_Loc item@(Item s _ _ _ ) = Loc s item
 
 instance Eq (StorageLocation t) where
   Loc s1 i1 == Loc s2 i2 = eqS s1 i1 s2 i2
--- a/src/Syntax/TimeAgnostic.hs
+++ b/src/Syntax/TimeAgnostic.hs
@@ -519,8 +519,15 @@ eval e = case e of
   ByStr _ s     -> pure . fromString $ s
   ByLit _ s     -> pure s
 
-  -- Eq  _ s a b     -> [a' == b' | a' <- eval a, b' <- eval b]
-  -- NEq _ s a b     -> [a' /= b' | a' <- eval a, b' <- eval b]
+  -- TODO better way to write these?
+  Eq _ SInteger x y -> [ x' == y' | x' <- eval x, y' <- eval y]
+  Eq _ SBoolean x y -> [ x' == y' | x' <- eval x, y' <- eval y]
+  Eq _ SByteStr x y -> [ x' == y' | x' <- eval x, y' <- eval y]
+
+  NEq _ SInteger x y -> [ x' /= y' | x' <- eval x, y' <- eval y]
+  NEq _ SBoolean x y -> [ x' /= y' | x' <- eval x, y' <- eval y]
+  NEq _ SByteStr x y -> [ x' /= y' | x' <- eval x, y' <- eval y]
+
   ITE _ a b c   -> eval a >>= \cond -> if cond then eval b else eval c
   _             -> empty
 
--- a/src/Syntax/TimeAgnostic.hs
+++ b/src/Syntax/TimeAgnostic.hs
@@ -256,7 +256,7 @@ instance Eq (Exp a t) where
   ByEnv _ a == ByEnv _ b = a == b
 
   Eq _ s a b == Eq _ s' c d = eqS s a s' c && eqS s b s' d
-  NEq _ s a b == Eq _ s' c d = eqS s a s' c && eqS s b s' d
+  NEq _ s a b == NEq _ s' c d = eqS s a s' c && eqS s b s' d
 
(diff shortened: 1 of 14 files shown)


src/Syntax/TimeAgnostic.hs
```diff
@@ -133,99 +131,97 @@ data StorageUpdate (t :: Timing) where
...
+  Slice :: Pn -> Exp AByteStr t -> Exp AInteger t -> Exp AInteger t -> Exp AByteStr t
+  ByStr :: Pn -> String -> Exp AByteStr t
+  ByLit :: Pn -> ByteString -> Exp AByteStr t
+  ByEnv :: Pn -> EthEnv -> Exp AByteStr t
 
   -- polymorphic
-  Eq  :: (Eq a, Typeable a) => Pn -> Exp a t -> Exp a t -> Exp Bool t
-  NEq :: (Eq a, Typeable a) => Pn -> Exp a t -> Exp a t -> Exp Bool t
-  ITE :: Pn -> Exp Bool t -> Exp a t -> Exp a t -> Exp a t
+  Eq  :: Pn -> SType a -> Exp a t -> Exp a t -> Exp ABoolean t
```
d-xo, line 217: Really slick, ended up making a lot of stuff neater

src/Syntax/TimeAgnostic.hs
```diff
@@ -489,8 +476,15 @@ eval e = case e of
   ByStr _ s     -> pure . fromString $ s
   ByLit _ s     -> pure s
 
-  Eq  _ a b     -> [a' == b' | a' <- eval a, b' <- eval b]
-  NEq _ a b     -> [a' /= b' | a' <- eval a, b' <- eval b]
+  -- TODO better way to write these?
+  Eq _ SInteger x y -> [ x' == y' | x' <- eval x, y' <- eval y]
```
d-xo, line 480: Out of curiosity, why doesn't the following work?

```suggestion
  Eq _ _ x y -> [ x' == y' | x' <- eval x, y' <- eval y]
```

src/Syntax/Types.hs
```diff
@@ -40,119 +45,80 @@ instance Show (SType a) where
...
--- We need this to be able to use 'SomeSing' when implementing 'ActType'.
 
--- Note that even though there only exist three different 'SType', this does
--- not mean that the type family is partial. It simply means that the resulting
--- type is uninhabited if the argument is neither 'Integer', 'Bool' nor
--- 'ByteString'.
-type instance Sing = SType
+-- -- | Compare equality of two things parametrized by types which have singletons.
+-- eqS :: forall (a :: *) (b :: *) f t. (SingI a, SingI b, Eq (f a t)) => f a t -> f b t -> Bool
+-- eqS fa fb = maybe False (\Refl -> fa == fb) $ testEquality (sing @a) (sing @b)
```
d-xo, line 59:
```suggestion
```

Is this needed?

src/Syntax/Types.hs
d-xo, line 1: This file becomes increasingly magical for me as it progresses hehe, but v happy to see it shrinking

src/Type.hs
```diff
@@ -204,7 +207,7 @@ noStorageRead store expr = for_ (keys store) $ \name ->
 -- ensures that key types match value types in an U.Assign
 checkAssign :: Env -> U.Assign -> Err [StorageUpdate]
 checkAssign env@Env{contract,store} (U.AssignVal (U.StorageVar _ (StorageValue (FromAbi typ)) name) expr)
-  = sequenceA [_Update (Item typ contract name []) <$> inferExpr env expr]
+  = sequenceA [_Update (_Item contract name []) <$> checkExpr env typ expr]
```
d-xo, line 210: I'm probably missing something obvious, but what is an `_Item` and where is it defined?

d-xo (commented):
Really nice cleanup, you're going to have to teach me how to use singletons properly now hehe. Just a couple of small questions and nits.

src/Syntax/TimeAgnostic.hs
zoep: It is not able to figure out that the type of `x'` and `y'` (which is `TypeOf a`)  have `Eq` instances. But once we make `a` concrete (e.g., `AInteger`) it knows that `TypeOf AInteger` is `Integer` and equality works.

Here, the singleton `SType` tag of `Eq`/`NEq` works as run-time witnesses of the `AType` parameter. So if we pattern match on `SInteger` we know that the `a` parameter of the expressions is `AInteger`.   I figured this while writing this code, and I though it was a very cool use of singleton types.

src/Syntax/Types.hs
zoep: It was the old, and more concise definition of `eqS`. I'm bringing this back, thanks for noticing.

src/Type.hs
zoep: These underscore definitions are a shorthand for some language constructors (`Item`, `Loc`, `TExp`, and others) that automatically retrieve the `SType` (here `typ`) when there is a `SingI` instance, and the user does not have to provide it. 

(These already existed before the PR, I just changed this line to use `_Item`, as in other places.)

https://github.com/ethereum/act/blob/227f74a73c0897a2a84600bebc2a81127ae764c1/src/Syntax/TimeAgnostic.hs#L163-L164


Here, because a is an `AType`, it has a corresponding singleton, which is retrieved by the typeclass method `sing`:

https://github.com/ethereum/act/blob/227f74a73c0897a2a84600bebc2a81127ae764c1/src/Syntax/Types.hs#L64-L66

PR #140, opened 2023-04-04
No filtering out variables from environment

zoep:
Removes filtering out variables from the environment that do not appear at the LHS of storage update declarations.

The main reason for this is that the strategy for removing variables with contract types was wrong, so it is a prerequisite for an upcoming PR that add support for contract types. 

Note that now we lift the restriction that storage variables that are used in storage update declaration have to be explicitly mentioned in the storage updated declarations.

### DIFF
--- a/doc/src/language.md
+++ b/doc/src/language.md
@@ -233,23 +233,3 @@ ensures
 
    x == _x
 ```
-
-## Referencing Storage Variables
-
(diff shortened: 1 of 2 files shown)


d-xo:
Can we update the docs? There is a section `Referencing Storage Variables` that can be removed now

d-xo (approved):
Looks good!

PR #141, opened 2023-04-05
Support for multiple contracts

zoep:
This adds parsing and type checking support for multiple contracts. In particular, it adds

 - parsing of constructor calls and accessing fields of contract storage variables 
 - a new contract type for Act expressions
 - typechecking of field accessing and constructor calls

The most significant AST change is a new syntactic category for storage references that can be either bare variables, map indices of storage references, or field accessing.  Roughly: 

`Entry := Var id |  Mapping Entry [Expr] | Field Entry id`

This simplifies representation and typechecking since it allows us to mix field accessing and map indices (e.g., `x.y[a].z`).


src/test/Test.hs
```diff
@@ -224,6 +224,7 @@ selectName typ (Names ints bools bytes) = do
                 AInteger -> ints
                 ABoolean -> bools
                 AByteStr -> bytes
+                AContract -> error "unsupported type"
```
d-xo, line 227: Is there any reason why we can't extend the generators to produce expressions involving contract types too?

src/Syntax/TimeAgnostic.hs
```diff
@@ -212,7 +228,8 @@ data Exp (a :: ActType) (t :: Timing) where
   ByStr :: Pn -> String -> Exp AByteStr t
   ByLit :: Pn -> ByteString -> Exp AByteStr t
   ByEnv :: Pn -> EthEnv -> Exp AByteStr t
-
+  -- contracts
+  Call   :: Pn -> SType a -> Id -> [TypedExp t] -> Exp a t
```
d-xo, line 232: I'm wondering if this should be `Create` instead of `Call`, since we use it only for contract creation, and we don't really have a notion of specifying actual calls within act right now?

tests/invariants/pass/amm-full.act
```diff
@@ -5,16 +5,12 @@ creates
...
 
 creates
 
-    Token token0 := t0
-    Token token1 := t1
-
-invariants
-
-    pre(token0.balanceOf[THIS]) * pre(token1.balanceOf[THIS]) <= post(token0.balanceOf[THIS]) * post(token1.balanceOf[THIS])
+    Token token0 := Token(s0)
```
d-xo, line 12: This is very close to the solidity syntax for casting an addressinto a contract type, and would read that way to may solidity developers IMO. What do you think about introducing a `create` keyword to make the construction of a new contract explicit here?

d-xo:
I don' t think it's a problem yet since we don't support contract types in constructor interfaces (?), but once it becomes possible to call a constructor with the result of another constructor, we're going to have to deal with orderings and cyclic reference detection. I also don't really view it as a problem in of itself, but the introduction of an ordering relation between contract contrsuctor calls will be a non trivial change to the philosophy of the language (which until now has had no notion of ordering at all).

src/Type.hs
```diff
@@ -221,69 +257,84 @@ checkAssign _ _ = error "todo: support struct assignment in constructors"
...
       , store    = Map.findWithDefault mempty name theirs
       }
 
+checkEntry :: forall t. Typeable t => Env -> U.Entry -> Err (SlotType, StorageRef t)
+checkEntry Env{contract,store,calldata} (U.EVar p name) = case (Map.lookup name store, Map.lookup name calldata) of
+  (Just _, Just _) -> throw (p, "Ambiguous variable " <> name)
+  (Just typ, Nothing) -> pure (typ, SVar p contract name)
+  (Nothing, Just _) -> throw (p, "Variable " <> name <> " is not a storage variable")
+  (Nothing, Nothing) -> throw (p, "Unknown storage variable " <> show name)
+  -- TODO more consitent check of name overlap between calldata and storage
```
d-xo, line 290: wdym here?

src/test/Test.hs
zoep: `genType` is only used to generate types of arguments, and we don't allow contract types where. This case will never actually match in the code since we never call `genType` with contract. I will add a more clear error message though.

src/Syntax/TimeAgnostic.hs
zoep: Yes, I think that's a great idea!

tests/invariants/pass/amm-full.act
zoep: I think that's great, I will add a new keyword

src/Type.hs
zoep: It refers to the line above: 
https://github.com/ethereum/act/blob/26525a8b8c11293b34dc7a1e05b2e46ca3361658/src/Type.hs#L286  

I think it would be better to check whether there are argument names that overlap with state variables in a different pass, rather than having this check whenever we lookup a name since there isn't a way to disambiguate the reference. wdyt?

src/Syntax/TimeAgnostic.hs
zoep: Done!

tests/invariants/pass/amm-full.act
zoep: Done!

PR #143, opened 2023-04-26
Change AST structure of Act specs

zoep:
This changes the untyped and typed ASTs of Act so that: 

- Contract structure is explicit: Ast specs are now a list of contracts. Each of them has exactly one constructor and a list of behaviours. 
- Constructors are explicit in the AST: implicit constructors are now made explicit in the AST during parsing. 
- Invariants are now part of constructors in the typed AST so that constructor arguments are in scope.


doc/src/language.md
```diff
@@ -241,25 +241,3 @@ iff
 storage
    a.x => z
 ```
-
```
zoep, line 244: This is a section in the docs that was removed by a PR because we no longer have this constraint. Then it was readded by a subsequent PR because of a merge error.

src/Syntax/TimeAgnostic.hs
```diff
@@ -50,22 +50,13 @@ import Syntax.Timing  as Syntax.TimeAgnostic
...
 
-
--- TODO maybe enforce this structure to Act programs
--- Store * [Contract of (C, [B], [I],)]
-
-data Transition t
-  = Ctor (Constructor t)
-  | Behv (Behaviour t)
-  deriving (Show, Eq)
+data Contract t = Contract [Constructor t] [Behaviour t]
```
d-xo, line 57: why is this a list of constructors?

zoep: Because in the typed AST we're splitting constructors into pass and fail modes, so we will have two constructors for each contract. But in a branch that builds on top of this one, I remove modes, and this is not a list anymore.

d-xo (approved):
looks good!

PR #144, opened 2023-04-28
Cycle detection in constructor call graphs

zoep:
Implements a depth first search in the constructor call graph to find circular dependencies, and throws an error if one is found.

### DIFF
--- a/src/Syntax.hs
+++ b/src/Syntax.hs
@@ -16,6 +16,7 @@ import Data.Map (Map,empty,insertWith,unionsWith,unionWith,singleton)
 
 import Syntax.TimeAgnostic as Agnostic
 import qualified Syntax.Annotated as Annotated
+import qualified Syntax.Typed as Typed
 import           Syntax.Untyped hiding (Constant,Rewrite,Contract)
 import qualified Syntax.Untyped as Untyped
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -71,17 +71,17 @@ test-cabal: src/*.hs
 
 # Just checks parsing
 tests/%.parse.pass:
-	./bin/act parse --file tests/$* > tests/$*.parsed.hs.out
-	diff tests/$*.parsed.hs.out tests/$*.parsed.hs
-	rm tests/$*.parsed.hs.out
```
d-xo, line 76: think you forgot to uncomment this?

src/Syntax.hs
```diff
@@ -115,6 +116,78 @@ locsFromExp = nub . go
       TEntry _ _ a -> locsFromItem a
       Var {} -> []
 
+callsFromExp :: Exp a t -> [Id]
```
d-xo, line 119: `createsFromExp` instead maybe?

Makefile
zoep: fixed!

src/Syntax.hs
zoep: fixed!

d-xo (approved):
looks good!

PR #145, opened 2023-04-28
Remove pass/fail modes from constructors and behaviours 

zoep:
This removes pass and fail modes from constructors and behaviours. It adds a new field in behaviour record to store the preconditions that come from case analysis (*) to semantically differentiate with behaviour preconditions that lead to revert. This is necessary to recover all the information that was stored in the fail-mode behaviour.

Most changes are trivial. One non-trivial one is the checking the equivalence with hevm revert paths.

(*) To ensure soundness the cases must still be exhaustive. A very simple way to guarantee that is by syntactically forcing the last case to be a wildcard, though we can opt for more sophisticated ways that involve asking the SMT.


src/Syntax/TimeAgnostic.hs
```diff
@@ -106,21 +105,15 @@ deriving instance Eq   (InvariantPred t) => Eq   (Constructor t)
 
 data Behaviour t = Behaviour
   { _name :: Id
-  , _mode :: Mode
   , _contract :: Id
   , _interface :: Interface
-  , _preconditions :: [Exp ABoolean t]
+  , _preconditions :: [Exp ABoolean t] -- if preconditions are not satisfied execution is reverted
+  , _case :: [Exp ABoolean t] -- if preconditions are satisfied and a case is not, some other instance of the bahaviour should apply
```
d-xo, line 111: I'm fine with either name if you feel strongly, but do you think that something like `_assumptions` might be clearer here?

src/Type.hs
```diff
@@ -196,31 +194,24 @@ checkTransition env (U.Transition _ name contract iface@(Interface _ decls) iffs
           negation = U.ENot nowhere $
                         foldl (\acc (U.Case _ e _) -> U.EOr nowhere e acc) (U.BoolLit nowhere False) rest
         in rest `snoc` (if isWild lastCase then U.Case pn negation post else lastCase)
+    -- TODO ensure non-overlapping and exhaustiveness (maybe with elaboration and mandatory wildcard?) 
 
     -- | split case into pass and fail case
```
d-xo, line 199: this comment is outdated...

zoep: I believe we still need to ensure that case branches are non-overlapping and the case analysis exhaustive for things to be sound. 

Non-overlapping is pretty easy to do, if we assume that a branch applies only if the one above does not hold, so I'll implement this for now, and leave an note for exhaustiveness.

PR #146, opened 2023-05-15
Multiple contracts for Coq backend

zoep:
This adds support for multiple contracts in the Coq backend. More specifically:

- Use Module to separate the code of each contract in Coq and provide separate namespace.

- Allow initialisation and update of storage variables that have contract types, constructor calls, and field accessing and update. 

- Add additional information to field accessors in the AST to denote the contract that the storage variable belongs too.

- Sort contracts topologically during typing so that there are no forward references to constructor calls that would cause an error in the Coq code.


src/Type.hs
```diff
@@ -96,35 +98,37 @@ typecheck' (U.Main contracts) = Act store <$> traverse (checkContract store cons
...
 
     calls = fmap findCalls $ contracts
     g = Map.fromList calls
-    
-    findCalls :: Contract -> (Id, [Id])
-    findCalls c@(Contract (Constructor cname _ _ _ _ _ _) _) = (cname, createsFromContract c)
-  
+
+    -- map a contract name to the list of contracts that it calls and its code
+    findCalls :: Contract -> (Id, ([Id], Contract))
```
d-xo, line 129: `findCreates`?

src/Type.hs
```diff
@@ -96,35 +98,37 @@ typecheck' (U.Main contracts) = Act store <$> traverse (checkContract store cons
...
 
 
-detectCycle :: Act -> Err ()
-detectCycle (Act _ contracts) =
-  () <$ foldValidation doDFS Set.empty (map fst calls)
+-- | Sort contracts topologically so there are no backward edges in
+-- the constructor call graph. Throw an error if a cycle is detected.
+topologicalSort :: Act -> Err Act
+topologicalSort (Act store contracts) =
+  Act store . reverse . map snd . OM.assocs <$> foldValidation doDFS OM.empty (map fst calls)
```
d-xo, line 105: maybe just a little note somewhere that a reverse post ordering of the vertices visited by the DFS is the same as a topological sort? this was a little non obvious to me at first

d-xo: super clean implementation though!

d-xo (approved):
Looks great, just a few small nits / comments

src/Type.hs
zoep: done!

src/Type.hs
zoep: done!

PR #148, opened 2023-05-18
Fix bug related to missing constraints from case conditions.

zoep:
This adds case conditions to the invariant constraints so that they are taken into account when checking invariants. This was a bug introduced by #145, that split preconditions into preconditions (i.e., ensures) and case conditions. 

It also adds a test that checks whether the case conditions are taken into account when proving invariants.


d-xo (approved):
looks good!

PR #150, opened 2023-05-31
TimeAgnostic to exposed modules

andrevidela:
For open games I've needed to expose the time-agnostic AST. Can we add it to the list of exported modules?


d-xo (approved):
looks good

PR #152, opened 2023-06-21
Equivalence with HEVM

zoep:
This PR implements a first prototype for checking equivalence between an Act spec and HEVM Expr obtained from EVM byte code. It works by translating Act to Expr and using the equivalence checker of HEVM to check equivalence. This is described in more detail in the [documentation](https://github.com/ethereum/act/blob/7f6e3b9a9d6a27bda5acee41d9d40f05ca55f31a/doc/src/equiv.md).

In more details the PR includes:

- The translation of Act to Expr (in [HEVM.hs](https://github.com/ethereum/act/blob/hevm-equiv/src/HEVM.hs)). 
- A top-level functionality that checks whether some Solidity code implements an Act spec (in [CLI.hs](https://github.com/ethereum/act/blob/hevm-equiv/src/CLI.hs)). 
- Implementation of the input space check (in [HEVM.hs](https://github.com/ethereum/act/blob/hevm-equiv/src/HEVM.hs))
- New syntax for in range assertions. When translating Act integers to machine words we need to implement integer bound differently. We add an `InRange` predicate to the typed AST that allows us to handle such bounds more modularly. We also need to be able to write conditional in-range assertions. So in the Act source instead of writing 
   ```
   iff in range uint256
     (a + b) + c
   iff 
     CALLVALUE == 0
   ```
  we now write 
   ```
   iff 
     CALLVALUE == 0
     inRange(uint256, (a + b) + c)
   ```
   In the typed AST, the in-range assertion gets elaborated to `InRange(a), InRange(b), InRange(c), InRange(a+b), InRange((a+b) +c)`, requiring that all intermediate expressions are within the given range and no overflow occurs. 
   
   Then, `InRange` predicates are translated according to the integer representation of each backend. 
   
- A new type for  `Store` that also keeps track of the order in which storage variables are declared. This is useful in order to provide a faithful storage layout when translating to Expr.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,10 +18,10 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
-      - uses: cachix/install-nix-action@6ed004b9ccb68dbc28e7c85bee15fa93dbd214ac # v22
+      - uses: cachix/install-nix-action@v22
         with:
           skip_adding_nixpkgs_channel: false
-      - uses: cachix/cachix-action@6a9a34cdd93d0ae4b4b59fd678660efb08109f2f # v12
+      - uses: cachix/cachix-action@v12
         with:
           name: dapp
           skipPush: true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
-      - uses: cachix/install-nix-action@29bd9290ef037a3ecbdafe83cbd2185e9dd0fa0a # v20
+      - uses: cachix/install-nix-action@6ed004b9ccb68dbc28e7c85bee15fa93dbd214ac # v20
         with:
           skip_adding_nixpkgs_channel: false
       - uses: cachix/cachix-action@6a9a34cdd93d0ae4b4b59fd678660efb08109f2f # v12
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
-      - uses: cachix/install-nix-action@6ed004b9ccb68dbc28e7c85bee15fa93dbd214ac # v20
+      - uses: cachix/install-nix-action@6ed004b9ccb68dbc28e7c85bee15fa93dbd214ac # v22
         with:
           skip_adding_nixpkgs_channel: false
       - uses: cachix/cachix-action@6a9a34cdd93d0ae4b4b59fd678660efb08109f2f # v12
--- a/doc/src/equiv.md
+++ b/doc/src/equiv.md
@@ -1,0 +1,50 @@
+# Checking equivalence with EVM bytecode
+
+EVM bytecode can be formally verified to implement an Act spec. This
+means that each successful behavior of the bytecode should be covered
+by the Act spect. To check equivalence the Act spec is translated to
+Expr, the intermediate representation of HEVM, and the EVM bytecode is
+symbolically executed to obtain its Expr representation. Then
+equivalence can be checked with the equivalence checker of HEVM.
+
+The Expr representation of an Act program is a list of *Success*
+nodes, that contain the possible successful results of the
+computation. The Expr representation of the EVM bytecode can also be
+flattened to a list of result nodes from which we only keep the
+successful executions, filtering out failed and partial execution
+paths and
+
+A success node in Expr, `Success cond res storage`, is a leaf in the
+Expr tree representation and contains the path conditions, `cond` that
+lead to the leaf, the result buffer `buf`, and the end state
+`storage`.
+
+
+## Equivalence checks 
+To check equivalence between the two Expr representations the
+following checks are performed. 
+
+### Result equivalence
+The two list of `Success` node are being checked for equivalence using
+the HEVM equivalence checker. For each pair of nodes in the two lists,
+we check that of all inputs that satisfy the path conditions the
+result and final storage the same. 
+
+### Input space equivalence
+Since the input space of the two list is not necessarily exhaustive,
+since some input may lead to failed execution paths that are not
+present in the list, we need to check that the input space of the two
+lists are the same. That is, there must not be inputs that satisfy
+some path condition in the first list but not the second and vice verse. 
+
+Say that the Act program has the Expr representation 
+`[Success c1 r1 s1, ..., Success cn rn sn`
+and the the EVM bytecode has the Expr representation 
+`[Success c1' r1' s1', ..., Success cn' rn' sn'`
+
+then we need to check that `c1 \/ .. \/ cn <-> c1' \/ .. \/ cn'` that
+is, we require that `c1 \/ .. \/ cn /\ ~ (c1' \/ .. \/ cn')` and `c1'
+\/ .. \/ cn' /\ ~ (c1 \/ .. \/ cn)` are both unsatisfiable.
+
+### Exhaustiveness checks for bytecode
+TODO
--- a/doc/src/equiv.md
+++ b/doc/src/equiv.md
@@ -12,7 +12,8 @@ nodes, that contain the possible successful results of the
 computation. The Expr representation of the EVM bytecode can also be
 flattened to a list of result nodes from which we only keep the
 successful executions, filtering out failed and partial execution
-paths.
+paths. An informative warning will be thrown when partial executions
+are encountered.
 
 A success node in Expr, `Success cond res storage`, is a leaf in the
 Expr tree representation and contains the path conditions, `cond` that
--- a/flake.nix
+++ b/flake.nix
@@ -5,7 +5,7 @@
     flake-utils.url = "github:numtide/flake-utils";
     nixpkgs.url = "github:nixos/nixpkgs";
     hevmUpstream = {
-      url = "github:ethereum/hevm/81f75a2ca34f5bb3202d207c3c07ca1740842717";
+      url = "github:ethereum/hevm/cd3440a26a7445f8acce84109db2179748c679aa";
       inputs.nixpkgs.follows = "nixpkgs";
     };
   };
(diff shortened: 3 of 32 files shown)


.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
-      - uses: cachix/install-nix-action@29bd9290ef037a3ecbdafe83cbd2185e9dd0fa0a # v20
+      - uses: cachix/install-nix-action@6ed004b9ccb68dbc28e7c85bee15fa93dbd214ac # v20
```
d-xo, line 21: can we just use a version tag here? (like `v22`)?

zoep: done!

src/CLI.hs
```diff
@@ -67,7 +79,8 @@ data Command w
                     }
 
   | HEVM            { spec       :: w ::: String               <?> "Path to spec"
-                    , soljson    :: w ::: String               <?> "Path to .sol.json"
+                    , sol        :: w ::: String               <?> "Path to .sol"
```
d-xo, line 82: I think the UX of taking a soliidty file instead of some build output is nice, but I think it's going to make it harder to integrate with real world projects which will have their own build system. Probably the nice way to do this is to offer some additional options that let us pull contracts from foundry build output. I'm fine if this gets done in a separate PR.

src/HEVM.hs
```diff
@@ -1,446 +1,365 @@
...
+import Syntax.Annotated
+import Syntax.Untyped (makeIface)
+import Syntax
 
-import EVM hiding (Query)
-import EVM.VMTest
-import EVM.Solidity hiding (Method)
-import EVM.ABI
-import EVM.Concrete
+import qualified EVM.Types as Types
```
d-xo, line 29:
```suggestion
import qualified EVM.Types as EVM
```

src/HEVM.hs
```diff
@@ -1,446 +1,365 @@
...
+import qualified Data.Text as T
+import qualified Data.Text.Lazy.IO as TL
+import qualified Data.ByteString.Char8 as B8 (pack)
+import Control.Concurrent.Async
 import Control.Monad
-import qualified Data.Vector as Vec
+import Data.DoubleWord
 
-import Print
+import Syntax.Annotated
```
d-xo, line 25:
```suggestion
import qualified Syntax.Annotated as Act
```

src/HEVM.hs
```diff
@@ -1,446 +1,365 @@
...
+    makeSig = T.pack $ makeIface iface
+    calldatas = fmap mkArg decls
+    (cdBuf, props) = combineFragments calldatas (Types.ConcreteBuf "")
+    withSelector = writeSelector cdBuf makeSig
+    sizeConstraints
+      = (bufLength withSelector Types..>= cdLen calldatas)
+        Types..&& (bufLength withSelector Types..< (Types.Lit (2 ^ (64 :: Integer))))
+  in (withSelector, sizeConstraints : props)
+
+makeCtrCalldata :: Interface -> Calldata
```
d-xo, line 71: whats the difference between this and `makeCalldata`?

src/HEVM.hs
```diff
@@ -1,446 +1,365 @@
...
+
+-- | Check whether two lists of behaviours cover exactly the same input space
+checkInputSpaces :: SolverGroup -> VeriOpts -> [Types.Expr Types.End] -> [Types.Expr Types.End] -> IO [EquivResult]
+checkInputSpaces solvers _ l1 l2 = do
+  let p1 = inputSpace l1
+  let p2 = inputSpace l2
+  let queries = fmap assertProps [ [ Types.PNeg (Types.por p1), Types.por p2 ]
+                               , [ Types.por p1, Types.PNeg (Types.por p2) ] ]
+
+  when True $ forM_ (zip [(1 :: Int)..] queries) $ \(idx, q) -> do
```
d-xo, line 317: `when True`?

d-xo (commented):
Looks awesome in general, mostly small comments and questions

doc/src/equiv.md
```diff
@@ -0,0 +1,50 @@
...
+by the Act spect. To check equivalence the Act spec is translated to
+Expr, the intermediate representation of HEVM, and the EVM bytecode is
+symbolically executed to obtain its Expr representation. Then
+equivalence can be checked with the equivalence checker of HEVM.
+
+The Expr representation of an Act program is a list of *Success*
+nodes, that contain the possible successful results of the
+computation. The Expr representation of the EVM bytecode can also be
+flattened to a list of result nodes from which we only keep the
+successful executions, filtering out failed and partial execution
```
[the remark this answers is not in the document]
zoep: Done and updated the doc

src/Print.hs
```diff
@@ -68,9 +68,10 @@ prettyExp e = case e of
   Mod _ a b -> print2 "%" a b
   Exp _ a b -> print2 "^" a b
   UIntMax _ a -> show $ uintmax a
-  UIntMin _ a -> show $ uintmin a
+  UIntMin _ a -> show $ uintmax a
```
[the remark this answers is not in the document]
zoep: typo!

src/HEVM.hs
zoep: Nice idea! What do you think about HEVM.Types? We could leave it unqualified since there are no name collisions anymore. Or maybe a better qualifier like HEVM? Right now the qualifier is `Types` and I don't like it at all.

src/HEVM.hs
zoep: yes!

zoep: or maybe `HEVM`

src/HEVM.hs
zoep: I think it is a nice idea! It will force us to write a lot of `Act.`s in the translation, but is more consistent with the way we import HEVM's ast. Also not sure..

src/HEVM.hs
zoep: yes it is a placefolder for the debug flag. The problem is that field selectors are not being exported.

src/CLI.hs
```diff
@@ -118,15 +130,22 @@ type' f = do
   contents <- readFile f
   validation (prettyErrs contents) (B.putStrLn . encode) (enrich <$> compile contents)
 
-prove :: FilePath -> Maybe Text -> Maybe Integer -> Bool -> IO ()
+-- type'' :: FilePath -> IO Act
+-- type'' f = do
+--   contents <- readFile f
+--   validation (\_ -> pure $ Act Map.empty []) pure (enrich <$> compile contents)
```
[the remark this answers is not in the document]
zoep: removed!

flake.nix
```diff
@@ -4,16 +4,25 @@
...
-        act = (pkgs.haskellPackages.callCabal2nixWithOptions "act" (gitignore ./src) "-fci" {})
+	myHaskellPackages = pkgs.haskellPackages.override {
+          overrides = self: super: rec {
+	    hevm = hevmUpstream.packages.${system}.noTests;
+          };
+	};
+        act = (myHaskellPackages.callCabal2nixWithOptions "act" (gitignore ./src) "-fci" {})
           .overrideAttrs (attrs : {
-            buildInputs = attrs.buildInputs ++ [ pkgs.z3 pkgs.cvc4 ];
+            buildInputs = attrs.buildInputs ++ [ pkgs.z3 pkgs.cvc4 pkgs.cvc5 ];
```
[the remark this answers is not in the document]
zoep: done!

flake.nix
```diff
@@ -22,14 +31,17 @@
...
+        devShell = with pkgs;
+          let libraryPath = "${lib.makeLibraryPath [ libff secp256k1 gmp ]}";
+          in myHaskellPackages.shellFor {
           packages = _: [ act ];
           buildInputs = with pkgs.haskellPackages; [
             cabal-install
             haskell-language-server
             pkgs.jq
             pkgs.z3
             pkgs.cvc4
```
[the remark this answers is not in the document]
zoep: done!

src/CLI.hs
```diff
@@ -41,6 +43,16 @@ import K hiding (normalize, indent)
 import SMT
 import Type
 import Coq hiding (indent)
+import HEVM
+
+import Debug.Trace
```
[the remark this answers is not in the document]
zoep: done!

src/HEVM.hs
zoep: The `makeCtrCalldata` does not have a function sector and it is intended to create the calldata for constructors. It is currently not used since we are not handling constructors yet. I do not remember the exact calling conventions for constructors but it wrote this after some conversations in the Berlin office. If we end up keeping this some things can be written more generically (e.g., `combineFragments`) for avoid code duplication between these two functions.

src/HEVM.hs
d-xo: either totally fine for me

src/HEVM.hs
d-xo: ah, we need to enable `OverloadedRecordDot` or smth in act too I guess to be able to get this nice modern dot notation for accessing record fields.

zoep: On nice! I did it locally for the file and now `opts.debug` works fine.

flake.nix
```diff
@@ -5,7 +5,7 @@
     flake-utils.url = "github:numtide/flake-utils";
     nixpkgs.url = "github:nixos/nixpkgs";
     hevmUpstream = {
-      url = "github:ethereum/hevm/e51f0bbdc6043aa2fb20cb69194215cdc62b5d75";
+      url = "github:ethereum/hevm/dbe984a96042a3873ecdffb84c60151d6229689a";
```
d-xo, line 8: can this be `main` now or are we still building against some `hevm` branch?

src/test/Test.hs
```diff
@@ -26,10 +26,11 @@ import Print (prettyBehaviour)
 import SMT
 import Syntax.Annotated
 
-import Debug.Trace
 import Text.Pretty.Simple
 import Data.Text.Lazy as T (unpack)
 
+import Debug.Trace
```
d-xo, line 32: can we remove this please

zoep: Actually it was already here because there are some tracing functions that use it at the end of the file. They are not used however, so we could delete them.

https://github.com/ethereum/act/blob/121353619ca740c0f2c02dd24cc156e09e557672/src/test/Test.hs#L276-L283

d-xo: ah, sorry didn't realise this was test code, think it's fine to leave them in

d-xo (approved):
looks good!

PR #154, opened 2023-07-07
ABI exhaustiveness check

zoep:
This PR includes:

- New check that checks whether all reachable successes in Solidity are described by the Act spec. This is checked by asserting that the function selector is not equal to the selectors of the interfaces in Act, and trying to find reachable successes.

- Removal of the range assertions in the initial calldata.

- Some cleanup and moving functions from CLI to HEVM.

- Some pretty colors for the output of Act.

### DIFF
--- a/src/CLI.hs
+++ b/src/CLI.hs
@@ -17,7 +17,6 @@ import System.Exit ( exitFailure )
 import System.IO (hPutStrLn, stderr, stdout)
 import Data.Text (pack, unpack)
 import Data.List
-import Data.Containers.ListUtils
 import Data.Maybe
 import Data.Traversable
 import qualified EVM.Solidity as Solidity
@@ -27,7 +26,6 @@ import qualified Data.Map.Strict as Map
 import Text.PrettyPrint.ANSI.Leijen hiding ((<$>))
 import GHC.Natural
 
-import qualified Data.ByteString as BS
 import qualified Data.ByteString.Lazy.Char8 as B
 
 import Control.Monad
@@ -49,10 +47,7 @@ import HEVM
 import EVM.SymExec
 import qualified EVM.Solvers as Solvers
 import EVM.Solidity
-import qualified EVM.Format as Format
 import qualified EVM.Types as EVM
-import qualified EVM.Expr as EVM
-import qualified EVM.Fetch as Fetch
 
 --command line options
 data Command w
@@ -226,7 +221,7 @@ hevm actspec cid sol' solver' timeout _ = do
   sequence_ $ flip fmap actbehvs $ \(name,behvs,calldata) ->
     Solvers.withSolvers solver' 1 (naturalFromInteger <$> timeout) $ \solvers -> do
       solbehvs <- removeFails <$> getBranches solvers bytecode calldata
-      putStrLn $ "\x1b[1mChecking behavior \x1b[4m" <> name <> " of Act\x1b[m"
+      putStrLn $ "\x1b[1mChecking behavior \x1b[4m" <> name <> "\x1b[m of Act\x1b[m"
       -- equivalence check
       putStrLn "\x1b[1mChecking if behaviour is matched by EVM\x1b[m"
       checkResult =<< checkEquiv solvers debugVeriOpts solbehvs behvs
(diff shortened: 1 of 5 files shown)


src/CLI.hs
```diff
@@ -223,114 +218,32 @@ hevm actspec cid sol' solver' timeout _ = do
...
     Solvers.withSolvers solver' 1 (naturalFromInteger <$> timeout) $ \solvers -> do
-    
       solbehvs <- removeFails <$> getBranches solvers bytecode calldata
+      putStrLn $ "\x1b[1mChecking behavior \x1b[4m" <> name <> "\x1b[m of Act\x1b[m"
       -- equivalence check
-      putStrLn "Checking if behaviours are equivalent"
-      checkResult =<< equivalenceCheck' solvers solbehvs behvs debugVeriOpts
-      -- exhaustiveness sheck
-      putStrLn "Checking if the input space is the same"
+      putStrLn "\x1b[1mChecking if behaviour is matched by EVM\x1b[m"
```
d-xo, line 226: We use this lib for pretty printing in the SMT output: https://hackage.haskell.org/package/ansi-wl-pprint-0.6.9/docs/Text-PrettyPrint-ANSI-Leijen.html which has a bunch of nice formatting and color combinators, it might be worth using here?

src/CLI.hs
```diff
@@ -223,114 +218,32 @@ hevm actspec cid sol' solver' timeout _ = do
...
+      checkResult =<< checkEquiv solvers debugVeriOpts solbehvs behvs
```
d-xo, line 227:
```suggestion
      checkResult =<< checkEquiv solvers defaultVeriOpts solbehvs behvs
```

src/CLI.hs
```diff
@@ -223,114 +218,32 @@ hevm actspec cid sol' solver' timeout _ = do
   let act = validation (\_ -> error "Too bad") id (enrich <$> compile specContents)
   bytecode <- fmap fromJust $ solcRuntime cid solContents
   let actbehvs = translateAct act
-  sequence_ $ flip fmap actbehvs $ \(behvs,calldata) ->
+  sequence_ $ flip fmap actbehvs $ \(name,behvs,calldata) ->
     Solvers.withSolvers solver' 1 (naturalFromInteger <$> timeout) $ \solvers -> do
```
d-xo, line 222: I think we can pull this out to the top level to avoid spawning and killing a solver on each run. We can also add some `-j` parameter or similar to control how many solvers we should spawn, and default that to the available number of cores on the system to support parallel solving.

src/HEVM.hs
```diff
@@ -365,3 +387,112 @@ checkOp (UIntMin _ _) = error "Internal error: invalid in range expression"
 checkOp (UIntMax _ _) = error "Internal error: invalid in range expression"
 checkOp (ITE _ _ _ _) = error "Internal error: invalid in range expression"
 checkOp (IntEnv _ _) = error "Internal error: invalid in range expression"
+
+
+-- | Checks whether all successful EVM behaviors are withing the
```
d-xo, line 392:
```suggestion
-- | Checks whether all successful EVM behaviors are within the
```

src/HEVM.hs
```diff
@@ -365,3 +387,112 @@ checkOp (UIntMin _ _) = error "Internal error: invalid in range expression"
...
+    checkBehv assertions (EVM.Success cnstr _ _ _) = assertions <> cnstr
+    checkBehv _ (EVM.Failure _ _ _) = []
+    checkBehv _ (EVM.Partial _ _ _) = []
+    checkBehv _ (EVM.ITE _ _ _) = error "Internal error: HEVM behaviours must be flattened"
+    checkBehv _ (EVM.GVar _) = error "Internal error: unepected GVar"
+
+    msg = "\x1b[1mThe following function selector results in behaviors not covered by the Act spec:\x1b[m"
+
+-- | decompiles the given EVM bytecode into a list of Expr branches
+getBranches :: SolverGroup -> BS.ByteString -> Calldata -> IO [EVM.Expr EVM.End]
```
d-xo, line 427: don't we have this in `SymExec.hs` in hevm already?

d-xo (approved):
Looks great! Just a few small comments

src/CLI.hs
zoep: I've already pulled that out in a subsequent PR (#156), where I only have one top-level `withSolvers`. I can also add a `-j` option.

src/HEVM.hs
zoep: Yes, we do. We can't directly use it because it's local to `checkEquiv` but at some point we should generalise it and use it here as well. Same for some other definitions in this file.

src/CLI.hs
zoep: It's fixed in #156! I will not change it here to avoid merge conflicts.

PR #155, opened 2023-07-10
HEVM equivalence CLI option + tests

zoep:
This PR

- Adds option to pass the runtime byte code to the CLI for hevm equivalence. 
- Enables the HEVM equivalence tests in the makefile


Makefile
```diff
@@ -43,7 +43,7 @@ postcondition_pass=$(wildcard tests/postconditions/pass/*.act) $(typing_pass)
 postcondition_fail=$(wildcard tests/postconditions/fail/*.act)
 
 # supposed to pass, but timeout
-hevm_buggy=tests/hevm/pass/safemath/safemath.act tests/hevm/pass/transfer/transfer.act
```
d-xo, line 46: do these still timeout?

d-xo (approved):
lgtm

Makefile
zoep: `Safemath` works, I've added back in #155. `transfer` I think we don't support yet.

PR #156, opened 2023-07-14
Constructor equivalence

zoep:
This PR add functionality for checking constructor equivalence. Equivalence checking follows the same principles as behaviour equivalence, with the difference that the init code is executed in hevm and the arguments are not passed with call date. 

There is also some code cleanup in the main hevm function in CLI.hs.


src/CLI.hs
```diff
@@ -183,42 +185,43 @@ coq' f = do
...
+      case (sol', code', initcode') of
+        (Just f, Nothing, Nothing) -> do
           solContents  <- TIO.readFile f
-          fmap fromJust $ solcRuntime cid solContents
-        (Nothing, Just c) -> pure $ Format.hexByteString "" c
-        (Nothing, Nothing) -> error "No EVM input is given. Please provide a Solidity file or EVM bytecode"
-        (Just _, Just _) -> error "Both Solidity file and bytecode are given. Please specify only one."
+          bytecodes cid solContents
+        (Nothing, Just c, Just i) -> pure (i, c)
+        (Nothing, Nothing, _) -> error "No runtime code is given"
```
d-xo, line 211: might be a bit nicer to print an error message and then exit cleanly with `exitFailure 1` here. It's not such a nice ux to dump a stack trace if the user forgets to pass some needed parameter

src/HEVM.hs
```diff
@@ -88,7 +87,7 @@ makeCtrCalldata (Interface _ decls) =
     mkArg :: Decl -> CalldataFragment
     mkArg (Decl typ x)  = symAbiArg (T.pack x) typ
     calldatas = fmap mkArg decls
-    (cdBuf, props) = combineFragments' calldatas 0 (EVM.AbstractBuf "txdata")
+    (cdBuf, props) = combineFragments' calldatas 0 (EVM.ConcreteBuf "")-- (EVM.AbstractBuf "txdata")
```
d-xo, line 90: why couldn't we use an `AbstractBuf` here again? would be cool to have some `TODO` comment explaining.

zoep: The code is trying to compute the lengths of the buffer and it fails with an unexpected symbolic argument. I don't know what we can do about it thought in the future..

src/CLI.hs
zoep: Done! Also removed an other similar error in the file.

Makefile
```diff
@@ -43,7 +43,7 @@ postcondition_pass=$(wildcard tests/postconditions/pass/*.act) $(typing_pass)
 postcondition_fail=$(wildcard tests/postconditions/fail/*.act)
 
 # supposed to pass, but timeout
-hevm_buggy=tests/hevm/pass/transfer/transfer.act tests/hevm/pass/simple/simple.act
+hevm_buggy=tests/hevm/pass/transfer/transfer.act tests/hevm/pass/simple/simple.act tests/hevm/pass/maps/maps.act
```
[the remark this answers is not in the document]
zoep: This is fixed now.

src/HEVM.hs
d-xo:
```suggestion
    -- We need to use a concrete buf as a base here because hevm bails when trying to execute with an abstract buf
    -- This is because hevm ends up trying to execute a codecopy with a symbolic size, which is unsupported atm
    -- This is probably unsound, but theres not a lot we can do about it at the moment...
    (cdBuf, props) = combineFragments' calldatas 0 (EVM.ConcreteBuf "")-- (EVM.AbstractBuf "txdata")
```

Maybe a comment like this?

d-xo (approved):
looks great!

PR #159, opened 2023-08-14
HEVM: start from concrete store for constructors

zoep:
Change the initial storage of constructors to an empty store instead of an abstract store.


d-xo (approved):
lgtm

PR #160, opened 2023-08-21
Port to symbolic addresses

zoep:
This PR adapts the hevm-Act equivalence checker to use symbolic addresses as the latest hevm. The port was mostly straightforward. Variable nodes in the typed AST were modified to carry their ABI type that was needed for the translation.


d-xo (approved):
lgtm!

PR #161, opened 2023-09-24
fix: crash on SPDX license identifiers

DennisDv24:
Currently, it looks like HEVM appends an "UNLICENSED" SPDX identifier to all contracts ([see](https://github.com/ethereum/hevm/blob/main/src/EVM/Solidity.hs#L667)). This makes the compilation process crash if you have your own identifiers in your contracts (`ParserError: Multiple SPDX license identifiers found in source file. Use \"AND\" or \"OR\" to combine multiple licenses. Please see https://spdx.org for more information.\n--> /tmp/nix-shell.075qFC/hevm62636-35.sol\n\n`) . I dont think this is the indended behaviour for Act, so I added a small regex to remove all SPDX identifiers before compiling anything.

### DIFF
--- a/src/CLI.hs
+++ b/src/CLI.hs
@@ -24,6 +24,7 @@ import qualified Data.Text as Text
 import qualified Data.Text.IO as TIO
 import Text.PrettyPrint.ANSI.Leijen hiding ((<$>))
 import GHC.Natural
+import Data.Function ((&))
 
 
 import qualified Data.ByteString.Lazy.Char8 as B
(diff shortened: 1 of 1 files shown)


d-xo:
Thanks for the report and the fix! I'm wondering if it might be cleaner to fix this in hevm directly, what do you think @zoep ?

zoep:
Thanks for the fix! Yes, I agree it would be cleaner and even more useful if we did that at hevm directly.

d-xo:
I updated the compilation helpers in hevm so that they no longer insert spdx headers and version bounds to the input source files. Closing this pr in favour of that one (https://github.com/ethereum/hevm/pull/387). 

Thanks again @DennisDv24 for the report and pr!

PR #162, opened 2023-09-28
Constructor calls in HEVM equivalence

zoep:
This PR adds support for multiple contracts in Act spec (WIP). More precisely:

- Allow constructor calls in constructors
- Remove `--contract` option, since all contracts in the spec are now checked by default

TODO: 
- Allow accessing external state variables in behaviours
- Add failing tests


Makefile
```diff
@@ -53,7 +53,7 @@ hevm_fail=$(wildcard tests/hevm/fail/*/*.act)
 failing_typing=tests/frontend/pass/array/array.act tests/frontend/pass/dss/vat.act tests/frontend/pass/creation/createMultiple.act tests/frontend/pass/staticstore/staticstore.act
 
 
-coq-examples = tests/coq/transitions tests/coq/safemath tests/coq/exponent tests/coq/token tests/coq/ERC20 tests/coq/multi
+coq-examples = tests/coq/transitions tests/coq/safemath tests/coq/exponent tests/coq/token tests/coq/ERC20 tests/coq/multi tests/coq/multi
```
zoep, line 56: This test should pass once we merge #388 into hevm.

src/CLI.hs
```diff
@@ -207,29 +207,32 @@ coq' f solver' smttimeout' debug' = do
...
+    cmap <- createContractMap contracts
+    let opts = if debug' then debugVeriOpts else defaultVeriOpts
 
+    Solvers.withSolvers solver' 1 (naturalFromInteger <$> timeout) $ \solvers ->
+      checkContracts solvers opts store cmap
   where
-    getBytecode :: IO (BS.ByteString, BS.ByteString)
-    getBytecode =
+
+    createContractMap :: [Contract] -> IO (Map Id (Contract, BS.ByteString, BS.ByteString))
```
zoep, line 221: We create a map from contact ids to their (Act spec, initcode, bytecode) triples. This makes it easy to look up the spec and the byte code of a contract when we encounter a call to its constructor. To check equivalence between the spec and the bytecodes, we traverse the map.

src/CLI.hs
```diff
@@ -207,29 +207,32 @@ coq' f solver' smttimeout' debug' = do
...
+
+    getBytecode :: Id -> IO (BS.ByteString, BS.ByteString)
+    getBytecode cid =
       case (sol', code', initcode') of
         (Just f, Nothing, Nothing) -> do
           solContents  <- TIO.readFile f
-          bytecodes cid solContents
-        (Nothing, Just c, Just i) -> pure (i, c)
+          bytecodes (Text.pack cid) solContents
+        (Nothing, Just _, Just _) -> render (text "Only Solidity file supported") >> exitFailure -- pure (i, c)
```
zoep, line 235: I'm not sure how we can support bytecode input when we have multiple contracts.

src/HEVM.hs
```diff
@@ -166,30 +185,137 @@ rewritesToExpr layout cid rewrites bytecode = foldl (flip $ rewriteToExpr layout
...
-rewriteToExpr :: Layout -> Id -> Rewrite -> ContractMap -> ContractMap
-rewriteToExpr _ _ (Constant _) cmap = cmap
-rewriteToExpr layout cid (Rewrite upd) cmap = updateToExpr layout cid upd cmap
+rewriteToExpr :: CodeMap -> Layout -> Id -> EVM.Expr EVM.EAddr -> Rewrite -> ContractMap -> ContractMap
+rewriteToExpr _ _ _ _ (Constant _) cmap = cmap
+rewriteToExpr codemap layout cid caddr (Rewrite upd) cmap = fst $ updateToExpr codemap layout cid caddr upd (cmap, [])
 
-updatesToExpr :: Layout -> Id -> [StorageUpdate] -> ContractMap -> ContractMap
-updatesToExpr layout cid upds initmap = foldl (flip $ updateToExpr layout cid) initmap upds
+updatesToExpr :: CodeMap -> Layout -> Id -> EVM.Expr EVM.EAddr -> [StorageUpdate] -> (ContractMap, [EVM.Prop]) -> (ContractMap, [EVM.Prop])
```
zoep, line 192: The return type propagates the preconditions of constructor calls upwards after substitution. As we discussed this is probably not needed. I will think about this a bit more and probably will be removing this from the return type.

zoep:
Subsumed by #167

PR #163, opened 2023-09-30
hie.yaml small rename

DennisDv24:
Im not sure what the HLS uses `hie.yaml` for, but I was getting this gignatic error from it that wouldnt let me use the LSP:
image
So I just renamed `act-internal` to `act` in `hie.yaml` and now its working smooth. Dk if this might break any of your envs tho.


d-xo:
Thanks!

PR #169, opened 2023-10-26
JSON nits for `Exp`

zoep:
Few small nits so that the JSON printer for `Exp` produces a more structured output.


sophierain (approved):
looks good

PR #172, opened 2023-11-09
Reorganise Directories and Module Structure

d-xo:
This reorganises the projects directory structure and module hierarchy a little to make it easier to consume externally:

1. act.cabal has been moved to the top level (now `cabal build` and friends work without having to `cd src` first
2. All modules have been prefixed with `Act`, making it more convenient to use as a library

I also moved the tests and executables into their own folders underneath `src` which makes the module structure a little clearer imo, and should make it easy to break our haskell tests up into separate modules.


zoep (approved):

zoep:
Looks good!

PR #173, opened 2023-11-14
Decompilation

d-xo:
Adds a certified decompiler that can automatically derive act specifications from evm bytecode. The process operates in the following stages:

1. Symbolically execute the bytecode to produce an EVM.Expr
2. Transform that Expr into one that can be safely represented using Integers
3. Convert that Expr into an Act spec (trusts solc compiler output)
4. Compile the generated Act spec back to Expr and check equivalence (solc compiler output no longer trusted)

This currently supports only very simple contracts, basically value types, simple arithmethic, and storage writes. Mappings in storage are currently not supported, but doing so should be in principle possible.

Some simple tests are added, but still todo is to hook this up to the main CLI interface.

Starting from the existing solidity code produces the following spec (I'm currently unsure where the `case True` comes from):

```solidity
contract C {
  uint x = 0;
  function f(uint v) public {
    x = v;
  }
}
```

```
constructor of C
interface C()

iff

  (CALLVALUE == 0)

creates

  uint256 x := 0


behaviour f of C
interface f(uint256 v)

iff

  (CALLVALUE == 0)

case

  true:

storage

  x => v
```


msooseth:
Hi!

Is this still being worked on? Should I review? Sounds cool :)

Mate

src/Act/HEVM.hs
```diff
@@ -1,20 +1,18 @@
...
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE OverloadedRecordDot #-}
-{-# LANGUAGE TypeApplications #-}
-{-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE NoFieldSelectors #-}
 
+-- {-# LANGUAGE TypeApplications #-}
```
d-xo, line 13:
```suggestion
```

src/Test/Decompile.hs
```diff
@@ -0,0 +1,124 @@
...
+import Act.Print
+import Act.CLI
+import Act.HEVM_utils
+
+import qualified EVM.Solvers as Solvers
+import EVM.Effects
+
+decompilerTests :: TestTree
+decompilerTests = testGroup "decompiler"
+  [ expectFail $ testCase "noop" $ checkDecompilation "C" [i|
```
d-xo, line 29: why is this one `expectFail`?

src/Act/Decompile.hs
```diff
@@ -0,0 +1,475 @@
...
+makeIntSafe solvers expr = evalStateT (mapExprM go expr) mempty
+  where
+    go :: forall b. EVM.Expr b -> StateT (Map (EVM.Expr EVM.EWord) EVM.Prop) m (EVM.Expr b)
+    go = \case
+      e@(EVM.Add a b) -> binop (EVM.Add a b .>= a) a b e
+      e@(EVM.Sub a b) -> binop (EVM.Sub a b .<= a) a b e
+      e@(EVM.Mul a b) -> binop (EVM.Div (EVM.Mul a b) b .== a) a b e
+      -- we can't encode symbolic exponents in smt, so we just always wrap
+      e@(EVM.Exp {}) -> pure (EVM.Mod e (EVM.Lit MAX_UINT))
+      -- TODO: I don't understand what this thing does, so just wrap it in a mod to be safe for now
```
d-xo, line 159: would be nice to figure this out I guess?

src/Act/Decompile.hs
```diff
@@ -0,0 +1,475 @@
...
+    layout <- ExceptT . pure . toErr "missing storage layout in solc output" $ contract.storageLayout
+    pure $ EVMContract
+      { name = snd $ T.breakOnEnd ":" contract.contractName
+      , storageLayout = layout
+      , runtime = behvs
+      , creation = ctor
+      }
+  where
+    creation = do
+      -- TODO: doesn't this have a 4 byte gap at the front?
```
d-xo, line 111: I guess this isn't true in the end?

src/Act/Decompile.hs
```diff
@@ -0,0 +1,475 @@
...
+
+-- | Transforms an EVM Expr into a form that it can be safely represented using
+-- an integer encoding. This basically makes all potential points of wrapping
+-- due to under/overflow explicit in the tree. This can certainly be made
+-- significantly more efficient and precise (abstract interpretation?), but
+-- this rather brute force attempt should be sound for now.
+makeIntSafe :: forall m a. App m => SolverGroup -> EVM.Expr a -> m (EVM.Expr a)
+makeIntSafe solvers expr = evalStateT (mapExprM go expr) mempty
+  where
+    go :: forall b. EVM.Expr b -> StateT (Map (EVM.Expr EVM.EWord) EVM.Prop) m (EVM.Expr b)
```
d-xo, line 152:
```suggestion
    -- walks the expression bottom up and checks (using an smt solver) if overflow is possible for any arithmetic operations it encounters. 
    -- results from child nodes are stored and reused as we move up the tree
    go :: forall b. EVM.Expr b -> StateT (Map (EVM.Expr EVM.EWord) EVM.Prop) m (EVM.Expr b)
```

src/Act/Decompile.hs
```diff
@@ -0,0 +1,475 @@
...
+        , _name = T.unpack method.name
+        , _preconditions = nub pres
+        , _caseconditions = mempty -- TODO: what to do here?
+        , _postconditions = mempty
+        , _stateUpdates = rewrites
+        , _returns = ret
+        }
+    mkbehv _ _ = error "Internal Error: mkbehv called on a non Success branch"
+
+-- TODO: we probably need to diff against the prestore or smth to be sound here
```
d-xo, line 267: Is this true? I don't really understand this comment.

d-xo (commented):
Thanks for fixing this up and getting it ready to merge. 
Mostly curious if some of the soundness concerns still in the `TODO` comments are valid.

msooseth:
Should I review? Is this active, do you want this merged? Looks like a lot of cool work!

src/Test/Decompile.hs
zoep: IIRC because we do not have a way to represent empty behaviours in Act

src/Act/Decompile.hs
```diff
@@ -0,0 +1,477 @@
...
+    -- Walks the expression bottom up and checks (using an smt solver) if overflow is possible for any 
+    -- arithmetic operations it encounters. Results from child nodes are stored and reused as we move up the tree
+    go :: forall b. EVM.Expr b -> StateT (Map (EVM.Expr EVM.EWord) EVM.Prop) m (EVM.Expr b)
+    go = \case
+      e@(EVM.Add a b) -> binop (EVM.Add a b .>= a) a b e
+      e@(EVM.Sub a b) -> binop (EVM.Sub a b .<= a) a b e
+      e@(EVM.Mul a b) -> binop (EVM.Div (EVM.Mul a b) b .== a) a b e
+      -- we can't encode symbolic exponents in smt, so we just always wrap
+      e@(EVM.Exp {}) -> pure (EVM.Mod e (EVM.Lit MAX_UINT))
+      -- TODO: I don't understand what this thing does, so just wrap it in a mod to be safe for now
```
d-xo, line 161:
```suggestion
      -- TODO: It's probably sound to remove the Mod here, but signextend is pretty mysterious so keeping it here for now to be on the safe side.
```

src/Test/Decompile.hs
d-xo: I think it's fine to leave that for another PR.

src/Act/Decompile.hs
d-xo: reviewed this section pretty carefully together with @zoep and we think this section is OK.
```suggestion
```

d-xo:
I'm not allowed to approve because I opened the PR originally, but all your changes look good to me @zoep and I'm fine to merge this whenever.

src/Test/Decompile.hs
zoep, line 29:
```suggestion
   -- TODO allow empty behaviours in source Act
  [ expectFail $ testCase "noop" $ checkDecompilation "C" [i|
```

PR #178, opened 2023-12-01
Proper threading of the Act state and some cleanup

zoep:
This PR includes:

- Properly threading the Act state between constructor and subsequent behaviours so that the counter for symbolic address names so that there are no name collisions if behaviours create new constructs. (Note: contract creation in behaviours has not been tested yet due to hevm returning only failed states.)
 
- Some general cleanup (removing comments, fixing typos)


d-xo (approved):
lgtm

PR #191, opened 2025-04-02
improve the readme for new users.

anjapetkovic:
I had some issues at first to set up act on my machine, so I updated the readme a bit to match the current state of the repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,9 @@ More in depth documentation can be found in [The Act Book](https://ethereum.gith
 
 # Building
 
-With nix:
+You can build the project with nix. If you do not have nix installed yet, you can try using the [Determinate Nix installer](https://github.com/DeterminateSystems/nix-installer).
+
+Building with nix:
 
 ```sh
 nix build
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,24 +16,25 @@ More in depth documentation can be found in [The Act Book](https://ethereum.gith
...
 
-Enter a nix-shell to get the dependencies of the project:
+After building, enter a nix-shell to get the dependencies of the project:
 
 ```sh
 nix develop
 ```
 
-you can then use `cabal` as normal:
+you can then use `cabal` as normal from where you have `act.cabal`, could be in `./src`:
```
zoep, line 35: `act.cabal` is at the root directory of act and it won't move (unless we move it).

README.md
```diff
@@ -42,11 +43,38 @@ to execute the unit tests:
...
+cabal v2-test # run haskell tests
 ```
 
 To update the project dependencies run:
 
 ```sh
 nix flake update
 ```
+
+# Usage
```
zoep, line 55: There is also a `--help` option for each one of them that describes all the available command line options.

README.md
```diff
@@ -42,11 +43,38 @@ to execute the unit tests:
...
+
+Once you are in the nix shell, you can use act backends as follows.
+
+## SMT
+
+```sh
+act prove --file <PATH_TO_SPEC>
```
zoep, line 62: Being able to use the `act` executable relies on having run `make` first, as it is the Makefile that runs the command to build it (of course one could also run it by hand).

I suggest using `cabal run act` instead which has no prerequisites, or specifying that `make` has to be run first.

zoep (changes requested):
Hi Anja, thanks for the contribution. I have added some comments.

While I think there is a lot of room of improvement in the README, this is better than what we had.

anjapetkovic:
The readme and the docs are now updated according to Zoe's suggestions.

PR #205, opened 2025-09-29
Support for HEVM equivalence of Vyper Contracts and multiple source files.

lefterislazar:
This PR adds support for Vyper contracts and proof of their equivalence through HEVM. To facilitate this, since Vyper allows for a single contract per file, support for multiple source files is added. Also added support for the Bitwuzla SMT theory solver, which demonstrated better performance and compatibility with the queries produced by HEVM during equivalence checking. 

## HEVM equivalence additions
Supporting  contracts produced by Vyper requires supporting Vyper's storage layout. There are 2 main differences with Solidity's layout that need to be accounted for.

1. Each storage element, whether a variable, a mapping slot or a mapping's element (or in future an array element), occupies a whole slot, regardless of type. This allows for the avoidance of bit masking techniques to extract a value from a slot. Instead, the whole slot is assumed to hold the value of the corresponding element.
For slots that hold integer values, this creates the requirement to satisfy the integer bounds of the corresponding integer type. This requirement is not enforced on the pre-state on a per-transition basis by the compiler, but holds only as an invariant across transitions. Thus, to prove equivalence, it is enough to assume the requirement for the pre-state, and given that, equivalence of behavior and input spaces ensures that the bounds hold on the post-state (as arithmetic in Vyper is checked).
Note that, in Solidity, assuming the integer bounds is not required, as the number of bits used in masking depends on each variable's size, and thus the bounds hold trivially. For now, it is still done as a sanity check.

2. When computing a mapping's element's slot, before application of the Keccak hash, the order of the concatenation of the mapping's slot index and the element's padded value, is reversed compared to Solidity's.

### Results
An implementation of ERC20 in Vyper has been added to the test suite, and is successfully proven equivalent to the specification used for the existing Solidity test. Use of Bitwuzla is recommended, as cvc5 presents compatibility issues, and Z3 presents performance issues.

## Multiple source files
When working with a single contract, Vyper compatibility is enabled by passing the single Vyper file to the new `--vy` argument instead of `--sol`.

To work with multiple source files, the `--sources` argument is added. It receives a .json file with the following structure:
- A `sources` object which contains a key for each relevant source file's path. The paths map to objects that contain a `language` key to specify between Solidity and Vyper (and compiled Bytecode in the future). Note that the file paths are relative to the json file's path.
- A `contracts` object which contains a key for each contract in the Act specification. The contracts map to objects that contain a `source` key, which maps to the file path of the source file that implements said contract.

The following is an example with 2 files and 2 contracts:
```
{
  "sources": {
    "file1.sol" : {
      "language" : "Solidity"
    },
    "file2.vy" : {
      "language" : "Vyper"
    }
  },
  "contracts" : {
    "A1" : {
      "source" : "file2.vy"
    },
    "A2" : {
      "source" : "file1.sol"
    }
  }
}
```

##Miscellaneous
- Small corrections in the integer bounds application logic.
- Regeneration of typing test results due to above changes.
- Addition of tests for Vyper contracts and mixed Vyper-Solidity multi-contract specifications.
- Set Bitwuzla as default solver for HEVM CI tests, due to better performance.


src/Act/CLI.hs
```diff
@@ -244,49 +252,140 @@ decompile' solFile' cid solver' timeout debug' = do
...
+    createContractMap :: [Contract] -> Map (Maybe Id) (LayoutMode, ByteString, ByteString) -> IO (Map Id (Contract, BS.ByteString, BS.ByteString), Map Id LayoutMode)
+    createContractMap contracts inputsMap | Map.keys inputsMap == [Nothing] =
+      -- Singleton inputsMap with Nothing as contract Id means that '--vy' was given
+      case contracts of
+        [spec'@(Contract cnstr _)] -> do
+          let cid =  _cname cnstr
+              (_, initcode'', runtimecode') = fromJust $ Map.lookup Nothing inputsMap
+          pure (Map.singleton cid (spec', initcode'', runtimecode'), Map.singleton cid VyperLayout)
+        _ -> do
+          render (yellow "Warning: " <> text "Multiple contracts in spec mapped to a single Vyper file" <> line)
```
zoep, line 280: We could also just throw an error here. It's very unlikely that the Act source will have multiple specs that match the same bytecode.

src/Act/HEVM.hs
```diff
@@ -434,32 +459,39 @@ expToBuf cmap styp e = do
...
+                  case layoutMode of
+                    SolidityLayout -> buf <> (wordToBuf slot')
+                    VyperLayout -> (wordToBuf slot') <> buf
+            pure (EVM.keccak concatenation)) slot ixs
+  pure (addr, EVM.Lit 0, size, layoutMode)
 refOffset _ (SField _ _ cid name) = do
   layout <- getLayout
-  let (slot, off, size) = getPosition layout cid name
-  pure (EVM.Lit (fromIntegral slot), EVM.Lit $ fromIntegral off, size)
+  let (slot, off, size, layoutMode) = getPosition layout cid name
```
zoep, line 493: trailing whitespace

src/Act/HEVM.hs
```diff
@@ -223,22 +247,23 @@ ifaceToSig (Interface name args) = Sig (T.pack name) (fmap fromdecl args)
...
-translateBehv :: Monad m => ContractMap -> [EVM.Prop] -> Behaviour -> ActT m (EVM.Expr EVM.End, ContractMap)
+translateBehv :: Monad m => ContractMap -> [EVM.Prop] -> Behaviour -> ActT m (EVM.Expr EVM.End, ContractMap, [EVM.Prop])
 translateBehv cmap cdataprops (Behaviour _ _ _ _ preconds caseconds _ upds ret)  = do
   preconds' <- mapM (toProp cmap) preconds
   caseconds' <- mapM (toProp cmap) caseconds
   ret' <- returnsToExpr cmap ret
   cmap' <- applyUpdates cmap cmap upds
   let acmap = abstractCmap initAddr cmap'
-  pure (EVM.Success (preconds' <> caseconds' <> cdataprops <> symAddrCnstr cmap') mempty ret' (M.map fst cmap'), acmap)
+  bounds <- storageBounds cmap $ nub $ concatMap locsFromExp preconds
```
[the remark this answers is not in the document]
lefterislazar: The explanation is the same as the one in the comment in `translateConstructor`. All locations that are active in a behaviour are present in `preconds` since they have all had integer bounds generated there. I will add a reference to the comment in said function.

src/Act/HEVM.hs
```diff
@@ -434,32 +459,39 @@ expToBuf cmap styp e = do
...
 -- | For the given storage reference, it returs the memory slot, the offset
 -- of the value within the slot, and the size of the value.
-refOffset :: Monad m => ContractMap -> Ref k -> ActT m (EVM.Expr EVM.EWord, EVM.Expr EVM.EWord, Int)
+refOffset :: Monad m => ContractMap -> Ref k -> ActT m (EVM.Expr EVM.EWord, EVM.Expr EVM.EWord, Int, LayoutMode)
 refOffset _ (CVar _ _ _) = error "Internal error: ill-typed entry"
 refOffset _ (SVar _ cid name) = do
   layout <- getLayout
-  let (slot, off, size) = getPosition layout cid name
-  pure (EVM.Lit (fromIntegral slot), EVM.Lit $ fromIntegral off, size)
+  let (slot, off, size, layoutMode) = getPosition layout cid name
```
zoep, line 476: Trailing whitespace

tests/hevm/pass/vyper/erc20/erc20.act
```diff
@@ -0,0 +1,214 @@
+constructor of Token
```
zoep, line 1: Awesome!

src/Act/CLI.hs
```diff
@@ -79,8 +84,10 @@ data Command w
 
   | HEVM            { spec       :: w ::: String               <?> "Path to spec"
                     , sol        :: w ::: Maybe String         <?> "Path to .sol"
+                    , vy         :: w ::: Maybe String         <?> "Path to .vy"
                     , code       :: w ::: Maybe ByteString     <?> "Runtime code"
                     , initcode   :: w ::: Maybe ByteString     <?> "Initial code"
+                    , sources    :: w ::: Maybe String         <?> "Path to sources .json"
```
zoep, line 90: How hard would it be to use this json file to allow writing Act contracts that span multiple files?

src/Act/CLI.hs
```diff
@@ -244,49 +252,140 @@ decompile' solFile' cid solver' timeout debug' = do
...
     case res of
       Success _ -> pure ()
       Failure err -> prettyErrs "" err
   where
-    createContractMap :: [Contract] -> IO (Map Id (Contract, BS.ByteString, BS.ByteString))
-    createContractMap contracts = do
-      foldM (\cmap spec'@(Contract cnstr _) -> do
+
+    -- Creates maps of storage layout modes and bytecodes, for all contracts contained in the given Act specification
+    createContractMap :: [Contract] -> Map (Maybe Id) (LayoutMode, ByteString, ByteString) -> IO (Map Id (Contract, BS.ByteString, BS.ByteString), Map Id LayoutMode)
```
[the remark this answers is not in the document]
zoep: It's not too bad having it separate from the beginning. I would only change it if merging the two maps leads to cleaner code.

zoep (approved):
Very nice work!

PR #227, opened 2026-07-29
fix: add missing Mod case to the typed-AST JSON emitter

mijovic:
`act type` crashed with `todo: json ast for: Mod ...` after successfully typechecking any spec that uses the % operator: the ToJSON instance for typed expressions covered every constructor except Mod, falling through to the error catch-all.

Add the Mod case (mirroring Div) and delete the catch-all, which the exhaustiveness checker now proves dead (Mod was the only missing constructor; keeping it trips -Werror=overlapping-patterns).

Regression test: a minimal %-using spec under tests/typing/pass, which exercises `act type | jq .` via the .type.pass make target. The existing %-specs under tests/hevm never hit this path.


lefterislazar:
Thank you for the fix.  I have added the missing test output. Once the CI passes we will accept and merge.

anjapetkovic (approved):
Looks good and the checks now pass.

PR #229, opened 2026-07-31
Fix error reporting for equivalence checking

lefterislazar:
This PR fixes error reporting for failures during equivalence checking, where source information is not required.
It also removes the now deprecated decompilation tests from the main test suite.


anjapetkovic (approved):
Looks good to me and the checks pass.

